PR #5, opened 2021-03-20
7902859: IntelliJ 2020.2: constructor must not have parameters: JTRegServiceConfigurable

cushon:
```
com.intellij.diagnostic.PluginException: Class constructor must not have parameters: com.oracle.plugin.jtreg.service.ui.JTRegServiceConfigurable [Plugin: jtreg]
	at com.intellij.serviceContainer.ComponentManagerImpl.instantiateExtensionWithPicoContainerOnlyIfNeeded(ComponentManagerImpl.kt:696)
	at com.intellij.openapi.options.ConfigurableEP$ClassProducer.createElement(ConfigurableEP.java:422)
	at com.intellij.openapi.options.ConfigurableEP.createConfigurable(ConfigurableEP.java:329)
	at com.intellij.openapi.options.ex.ConfigurableWrapper.createConfigurable(ConfigurableWrapper.java:45)
```

---------
### Progress
- [x] Referenced JBS issue must only be used for a single change

### Issue
 * [CODETOOLS-7902859](https://bugs.openjdk.java.net/browse/CODETOOLS-7902859): IntelliJ 2020.2: constructor must not have parameters: JTRegServiceConfigurable

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.java.net/census#mcimadamore) (@mcimadamore - Committer) ⚠️ Review applies to 1c58dbf6bd0032803e8083bb828e18fb486f1208

### Download
To checkout this PR locally:
`$ git fetch https://git.openjdk.java.net/jtreg pull/5/head:pull/5`
`$ git checkout pull/5`

To update a local copy of the PR:
`$ git checkout pull/5`
`$ git pull https://git.openjdk.java.net/jtreg pull/5/head`

### DIFF
--- a/plugins/idea/resources/META-INF/plugin.xml
+++ b/plugins/idea/resources/META-INF/plugin.xml
@@ -26,14 +26,14 @@
 <idea-plugin>
     <id>jtreg</id>
     <name>jtreg Test Support</name>
-    <version>1.11</version>
+    <version>1.12</version>
     <description><![CDATA[
       Allows execution of tests developed using the <a href="http://openjdk.java.net/jtreg/">jtreg</a> framework.
     ]]></description>
 
     <change-notes><![CDATA[
     <ul>
-      <li>Fix crash in IntelliJ IDE2020.1</li>
+      <li>Fix compatibility with 2020.2</li>
     </ul>
     ]]>
     </change-notes>
--- a/plugins/idea/src/com/oracle/plugin/jtreg/service/ui/JTRegServiceConfigurable.java
+++ b/plugins/idea/src/com/oracle/plugin/jtreg/service/ui/JTRegServiceConfigurable.java
@@ -73,15 +73,18 @@ public class JTRegServiceConfigurable implements SearchableConfigurable {
     private JPanel myListPane;
     private CollectionListModel<AntBuildTarget> myModel;
 
-
     Project project;
-    JTRegService service;
-    AntConfigurationBase antConfiguration;
 
-    public JTRegServiceConfigurable(Project project, JTRegService service, AntConfiguration antConfiguration) {
+    public JTRegServiceConfigurable(Project project) {
         this.project = project;
-        this.service = service;
-        this.antConfiguration = (AntConfigurationBase)antConfiguration;
+    }
+
+    private JTRegService getJTRegService() {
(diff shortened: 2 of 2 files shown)


mcimadamore (commented):
Looks good - typically we add notes of what changed in the plugin file (resources/plugin.xml) and we bump the version number.
Also, I suspect this fix has some implication when it comes to backward compatibility - currently we state:
`    <idea-version since-build="162.0"/>`
Should we rectify that?

mcimadamore:
Btw - this also needs a bug to be opened under the CodeTools project (tool = jtreg).

cushon:
Thanks!

I filed the bug, and updated the note and version number in `resources/plugin.xml`.

> Also, I suspect this fix has some implication when it comes to backward compatibility - currently we state: `<idea-version since-build="162.0"/>`

I tried downloading 162.0 / 2016.2, the I can still install the new plugin and load the preference pane without an error, so I think this change is backwards compatible. I don't know very much about IntelliJ plugins, though, and I can do more investigation if there's anything you'd recommend verifying? Alternately I can just bump the version in `<idea-version>` if you think everyone using the plugin is likely to be on a more recent version of IntelliJ at this point?

mcimadamore:
> Thanks!
> 
> I filed the bug, and updated the note and version number in `resources/plugin.xml`.
> 
> > Also, I suspect this fix has some implication when it comes to backward compatibility - currently we state: `<idea-version since-build="162.0"/>`
> 
> I tried downloading 162.0 / 2016.2, the I can still install the new plugin and load the preference pane without an error, so I think this change is backwards compatible. I don't know very much about IntelliJ plugins, though, and I can do more investigation if there's anything you'd recommend verifying? Alternately I can just bump the version in `<idea-version>` if you think everyone using the plugin is likely to be on a more recent version of IntelliJ at this point?

Thanks for trying out on 2016.2 - if the pane showed up, then I think we're ok. At some point yes, we'll need to bump. But we can do that separately.

mcimadamore (approved):
Looks good - there seems to be a jcheck issue?

cushon:
> there seems to be a jcheck issue

Fixed now, thanks. (I added the bug ID to both the commit message and the PR title, it looks like it's only supposed to be used for the PR title.)

PR #44, opened 2021-12-03
7903071: Jtreg plugin debug functionality broken in 2021.3

mcimadamore:
IntelliJ 2021.3 changed the order in which certain plugin methods are called when debugging a jtreg test. More specifically, the `JtregDebuggerRunner::createContentDescriptor` method is called _after_ `JTRegConfigurationRunnableState::createJavaParameters`. Since the former sets up the debugging port which has then to be passed to the jtreg command line (in the latter), this means that, in 2021.3 we end up creating a command line that is broken because it has either no port (when we debug for the first time) or a stale port (when we debug subsequent times).

I discussed this with Anna Kozlova and, while jtreg is a bit peculiar compared to other debugging plugin (as jtreg spawns other VMs, which is why we need to create a command line which includes the port number), it is possible to fix this in a way that is also compatible with older versions - by overriding `JtregDebuggerRunner::execute`, which is called before the other two methods. So we can setup the port number in there - and then reuse port number in the subsequent calls.

---------
### Progress

### Issue
 * [CODETOOLS-7903071](https://bugs.openjdk.java.net/browse/CODETOOLS-7903071): Jtreg plugin debug functionality broken in 2021.3

### Reviewers
 * [Jonathan Gibbons](https://openjdk.java.net/census#jjg) (@jonathan-gibbons - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jtreg pull/44/head:pull/44` \
`$ git checkout pull/44`

`$ git checkout pull/44` \
`$ git pull https://git.openjdk.java.net/jtreg pull/44/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 44`

`$ git pr show -t 44`

</details>

<a href="https://git.openjdk.java.net/jtreg/pull/44.diff">https://git.openjdk.java.net/jtreg/pull/44.diff</a>

</details>


jonathan-gibbons (approved):
I'm not an expert on this code, but there may not be (m)any other experts.

The code changes look reasonable.

LanceAndersen (commented):
Looks OK to me

PR #50, opened 2022-02-07
CODETOOLS-7903091: Convert jtreg to use NIO

jonathan-gibbons:
Please review a medium-big, conceptually simple but occasionally tricky update to jtreg, to use `java.nio.file.Path` instead of `java.io.File` where reasonable.

While most of the changes are formulaic and follow the obvious patterns, there were some noteworthy points:

* The most complicated class to change was `Locations`, which does the most work to derive new paths. In particular, the semantics of `new File(a, b)` do not always match `Path.of(a).resolve(b)` for corner cases like when `a` is`null` or empty, or when `b` is absolute.

* The NIO equivalent for some operations on `File` throw checked IO exceptions.  These methods were wrapped to to throw a custom unchecked IO exception.

* The NIO equivalent for creating a new file (`Path`) throws an unchecked `InvalidPathException` exception. The use sites are too numerous to handle individually.  The solution is to catch exceptions like these (and the ones described in the previous bullet) at strategic points where they can be handled or wrapped back into (existing) checked exceptions.

The development has been iterative, frequently running the jtreg internal tests locally (on a Mac) and occasionally building on a CI system (Linux).  We should do more exhaustive testing with JDK before this work ever gets promoted.

---------
### Progress

### Issue
 * [CODETOOLS-7903091](https://bugs.openjdk.java.net/browse/CODETOOLS-7903091): Convert jtreg to use NIO

### Reviewers
 * [Iris Clark](https://openjdk.java.net/census#iris) (@irisclark - **Reviewer**) ⚠️ Review applies to c2e545438f6d817113881242b0933e554d18e195

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jtreg pull/50/head:pull/50` \
`$ git checkout pull/50`

`$ git checkout pull/50` \
`$ git pull https://git.openjdk.java.net/jtreg pull/50/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 50`

`$ git pr show -t 50`

</details>

<a href="https://git.openjdk.java.net/jtreg/pull/50.diff">https://git.openjdk.java.net/jtreg/pull/50.diff</a>

</details>

### DIFF
--- a/src/share/classes/com/sun/javatest/regtest/tool/OptionDecoder.java
+++ b/src/share/classes/com/sun/javatest/regtest/tool/OptionDecoder.java
@@ -25,7 +25,6 @@
 
 package com.sun.javatest.regtest.tool;
 
-import java.io.File;
 import java.nio.file.InvalidPathException;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -192,10 +191,6 @@ public class OptionDecoder {
             throw new BadArgs(i18n, "opt.bad.path.for.option", arg, e.getInput(), e.getMessage());
         }
     }
-
-//    public void addFile(File file) throws BadArgs {
-//        fileOption.process(null, file.getPath());
-//    }
 
     public void addFile(String path) throws BadArgs {
         fileOption.process(null, path);
--- a/src/share/classes/com/sun/javatest/regtest/tool/OptionDecoder.java
+++ b/src/share/classes/com/sun/javatest/regtest/tool/OptionDecoder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2007, 2016, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2007, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -26,6 +26,7 @@
 package com.sun.javatest.regtest.tool;
 
 import java.io.File;
+import java.nio.file.InvalidPathException;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
@@ -185,12 +186,16 @@ public class OptionDecoder {
         if (debugOptions)
             System.err.println("OptionDecoder.decodeArg: " + name + " " + value);
 
-        o.process(arg, value);
+        try {
+            o.process(arg, value);
+        } catch (InvalidPathException e) {
+            throw new BadArgs(i18n, "opt.bad.path.for.option", arg, e.getInput(), e.getMessage());
+        }
     }
 
-    public void addFile(File file) throws BadArgs {
-        fileOption.process(null, file.getPath());
-    }
+//    public void addFile(File file) throws BadArgs {
+//        fileOption.process(null, file.getPath());
+//    }
 
     public void addFile(String path) throws BadArgs {
         fileOption.process(null, path);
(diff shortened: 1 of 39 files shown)


src/share/classes/com/sun/javatest/regtest/tool/OptionDecoder.java
```diff
@@ -185,12 +186,16 @@ private void decodeArg(String arg, Iterator<String> iter) throws BadArgs {
...
+            throw new BadArgs(i18n, "opt.bad.path.for.option", arg, e.getInput(), e.getMessage());
+        }
     }
 
-    public void addFile(File file) throws BadArgs {
-        fileOption.process(null, file.getPath());
-    }
+//    public void addFile(File file) throws BadArgs {
+//        fileOption.process(null, file.getPath());
+//    }
```
irisclark, line 198: Retained intentionally or to-be-converted?

src/share/classes/com/sun/javatest/regtest/util/FileUtils.java
```diff
@@ -0,0 +1,155 @@
+package com.sun.javatest.regtest.util;
```
irisclark, line 1: No legal notices (copyright or license)?

src/share/classes/com/sun/javatest/regtest/util/FileUtils.java
```diff
@@ -0,0 +1,155 @@
...
+                default:
+                    return "Cannot perform unknown operation for " + p;
+            }
+        }
+    }
+
+    private FileUtils() { }
+
+    /**
+     * Returns the size of a file, in  bytes.
```
irisclark, line 63: Extra space before "bytes", also at line 66.

src/share/classes/com/sun/javatest/regtest/util/FileUtils.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    /**
+     * Compares the times that a pair of files were last modified.
+     *
+     * @param p1 the first file
+     * @param p2 the second file
+     * @return {@code -1}, {@code 0} or {@code 1} according to whether the last modified time of the first file
+     *          is less than, equal to, or greater than the last modified time of the second file
+     *
```
irisclark, line 99: Not sure I understand the spacing convention, but the presence of this empty line and those at lines 109, 111, 113, and 128 may be inconsistent.

irisclark (commented):
I think this is a great first pass at modernizing jtreg!  I only had a few minor comments.

Due to the nature of this change, I'd be reluctant to recommend cutting over to this version of jtreg without wide communication and extensive testing on the existing test suite.  For testing, minimally, you should confirm that, for a given set of tests, you have the same number of pass/fail tests using both the old and new versions of the tool and the failures themselves are consistent (i.e. test failures occur because of the same problem in the system, not jtreg file/path handling).  Recommend that this test be repeated across multiple platforms.

src/share/classes/com/sun/javatest/regtest/tool/OptionDecoder.java
jonathan-gibbons: ... probably, to be removed

jonathan-gibbons: removed

src/share/classes/com/sun/javatest/regtest/util/FileUtils.java
jonathan-gibbons: Added

src/share/classes/com/sun/javatest/regtest/util/FileUtils.java
jonathan-gibbons: fixed

src/share/classes/com/sun/javatest/regtest/util/FileUtils.java
jonathan-gibbons: Inconsistency is bad, especially within a new file. I was aiming for a more spaced-out look, more similar to the core-libs style, with blank lines between groups of tags.

irisclark (approved):

PR #70, opened 2022-04-08
7903151: NPE when resolving @library from jtreg plugin

mcimadamore:
This patch fixes a trivial NPE which occurs when the plugin is attempting to resolve an absolute library in a `@library` tag. If no library is found starting from TEST.ROOT, and if TEST.ROOT does not specify any alternate library root, the plugin fails with a NPE, as the code is eagerly trimming the property value "external.lib.roots", which might not exist.

The solution is simple, e.g. check for null before trimming.

---------
### Progress

### Issue
 * [CODETOOLS-7903151](https://bugs.openjdk.java.net/browse/CODETOOLS-7903151): NPE when resolving @library from jtreg plugin

### Reviewers
 * [Daniel Jeliński](https://openjdk.java.net/census#djelinski) (@djelinski - no project role)
 * [Jonathan Gibbons](https://openjdk.java.net/census#jjg) (@jonathan-gibbons - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.java.net/jtreg pull/70/head:pull/70` \
`$ git checkout pull/70`

`$ git checkout pull/70` \
`$ git pull https://git.openjdk.java.net/jtreg pull/70/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 70`

`$ git pr show -t 70`

</details>

<a href="https://git.openjdk.java.net/jtreg/pull/70.diff">https://git.openjdk.java.net/jtreg/pull/70.diff</a>

</details>

### DIFF
--- a/plugins/idea/build.gradle
+++ b/plugins/idea/build.gradle
@@ -26,7 +26,7 @@ dependencies {
 
 // See https://github.com/JetBrains/gradle-intellij-plugin/
 intellij {
-    version = '2021.3.3'
+    version = minBuild
     plugins = ['java', 'AntSupport', 'TestNG-J']
 }
 
--- a/plugins/idea/build.gradle
+++ b/plugins/idea/build.gradle
@@ -26,6 +26,7 @@ dependencies {
 
 // See https://github.com/JetBrains/gradle-intellij-plugin/
 intellij {
+    version = '2021.3.3'
     plugins = ['java', 'AntSupport', 'TestNG-J']
 }
 
--- a/plugins/idea/README.md
+++ b/plugins/idea/README.md
@@ -38,7 +38,7 @@ This will download gradle and the required IntelliJ dependencies, will build the
 
 > Note: to build the plugin, the build script must point to a valid jtreg installation; see the property `jtregHome` in the `gradle.properties` file, and tweak accordingly.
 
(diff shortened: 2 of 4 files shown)


plugins/idea/build.gradle
```diff
@@ -26,6 +26,7 @@ dependencies {
 
 // See https://github.com/JetBrains/gradle-intellij-plugin/
 intellij {
+    version = '2021.3.3'
```
mcimadamore, line 29: This change to the build was necessary as the 1.0 of the IntelliJ gradle plugin seems to have an issue when working with the latest IntelliJ EAP (which the build downloads so that it can build against it). This problem is documented here:

https://youtrack.jetbrains.com/issue/IDEA-291192

The solution would be to update to plugin version 1.5.2 - but that plugin version still requires an intellij version to be specified manually. I found that by specify the current IntelliJ version as the IntelliJ API we wanna build against, things work just fine, even with the 1.0 plugin.

djelinski, line 29:
```suggestion
    version = project.findProperty('intellijVersion') ?: '2021.3.3'
```
while at it, you could also update the docs on `intellijVersion` [here](https://github.com/openjdk/jtreg/blob/master/plugins/idea/README.md?plain=1#L41)

jonathan-gibbons (approved):

PR #78, opened 2022-04-23
7903047: Make JUnit runner use the JUnit Platform Launcher

sormuras:
Make JUnit runner use the JUnit Platform Launcher

See https://bugs.openjdk.java.net/browse/CODETOOLS-7903047

### DIFF
--- a/src/share/classes/com/sun/javatest/regtest/agent/JUnitRunner.java
+++ b/src/share/classes/com/sun/javatest/regtest/agent/JUnitRunner.java
@@ -25,17 +25,21 @@
 
 package com.sun.javatest.regtest.agent;
 
+import org.junit.platform.engine.discovery.DiscoverySelectors;
+import org.junit.platform.launcher.core.LauncherDiscoveryRequestBuilder;
+import org.junit.platform.launcher.core.LauncherFactory;
+import org.junit.platform.launcher.listeners.SummaryGeneratingListener;
+
 import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.lang.reflect.Method;
-
 
 /**
  * TestRunner to run JUnit tests.
  */
 public class JUnitRunner implements MainActionHelper.TestRunner {
     private static final String
-        JUNIT_NO_DRIVER        = "No JUnit 4 driver (install junit.jar next to jtreg.jar)";
+        JUNIT_NO_DRIVER        = "No JUnit driver (install junit-platform-console-standalone.jar next to jtreg.jar)";
 
     public static void main(String... args) throws Exception {
         main(null, args);
@@ -69,6 +73,11 @@ public class JUnitRunner implements MainActionHelper.TestRunner {
             cl = JUnitRunner.class.getClassLoader();
         }
         Class<?> mainClass = Class.forName(className, false, cl);
+        // runWithJUnit4(mainClass);
+        runWithJUnitPlatform(mainClass);
+    }
+
+    private static void runWithJUnit4(Class<?> mainClass) throws Exception {
         org.junit.runner.Result result;
         try {
             result = org.junit.runner.JUnitCore.runClasses(mainClass);
@@ -91,4 +100,36 @@ public class JUnitRunner implements MainActionHelper.TestRunner {
         }
     }
 
+    private static void runWithJUnitPlatform(Class<?> mainClass) throws Exception {
+        // https://junit.org/junit5/docs/current/user-guide/#launcher-api-execution
+        Thread.currentThread().setContextClassLoader(mainClass.getClassLoader());
+        try {
+
+            var request = LauncherDiscoveryRequestBuilder.request()
+                .selectors(DiscoverySelectors.selectClass(mainClass))
+                .build();
+
+            var summaryGeneratingListener = new SummaryGeneratingListener();
+
+            try (var session = LauncherFactory.openSession()) {
+                var launcher = session.getLauncher();
+                launcher.registerTestExecutionListeners(summaryGeneratingListener);
+                launcher.execute(request);
+            }
+
+            var summary = summaryGeneratingListener.getSummary();
+            if (summary.getTotalFailureCount() > 0) {
+                var sw = new StringWriter();
+                try (var pw = new PrintWriter(sw)) {
+                    pw.println("JavaTest Message: JUnit Platform Failure(s): " + summary.getTotalFailureCount());
+                    summary.printTo(pw);
+                }
+                System.err.println(sw);
+                throw new Exception("JUnit test failure");
+            }
+
+        } catch (NoClassDefFoundError ex) {
+            throw new Exception(JUNIT_NO_DRIVER, ex);
+        }
+    }
 }


make/Defs.gmk
```diff
@@ -111,7 +111,7 @@ JDKJAVA = $(JDKHOME)/bin/java
 JDKJAVAC = $(JDKHOME)/bin/javac
 JAR = $(JDKHOME)/bin/jar
 
-AGENT_JAVAC_SOURCE_TARGET = --release 8
+AGENT_JAVAC_SOURCE_TARGET = --release 11
```
jonathan-gibbons, line 114: No. The use of 8 is deliberate, so that we can run use jtreg to test older versions of JDK.

jonathan-gibbons (changes requested):

make/Defs.gmk
sormuras: OK. Reverting to `--release 8` and replacing all `var` usages with explicit type references.

jonathan-gibbons: Maybe prefix the line with a comment of the form:
```
// Specify oldest supported JDK version that can be tested
```

src/share/classes/com/sun/javatest/regtest/agent/JUnitRunner.java
```diff
@@ -25,17 +25,27 @@
...
 import java.lang.reflect.Method;
 
-
 /**
  * TestRunner to run JUnit tests.
  */
 public class JUnitRunner implements MainActionHelper.TestRunner {
     private static final String
-        JUNIT_NO_DRIVER        = "No JUnit 4 driver (install junit.jar next to jtreg.jar)";
+        JUNIT_NO_DRIVER        = "No JUnit driver (install junit-platform-console-standalone.jar next to jtreg.jar)";
```
jonathan-gibbons, line 46: The code should not specifically require this jar file by name. We went to an effort recently to decouple the names of the jar files used to build jtreg from the code that uses them.
Admittedly, this string literal was an accidental oversight from that recent work. I suggest changing the name of the jar file to just `JUnit` or `JUnit jar file(s)`

src/share/classes/com/sun/javatest/regtest/agent/JUnitRunner.java
```diff
@@ -25,17 +25,27 @@
...
+    private static final boolean
+        JUNIT_RUN_WITH_JUNIT_4 = Flags.get("runWithJUnit4");
```
jonathan-gibbons, line 48: suggest adding comment that this is temporary, during the transition.

src/share/classes/com/sun/javatest/regtest/agent/JUnitRunner.java
```diff
@@ -91,4 +109,40 @@ public static void main(ClassLoader loader, String... args) throws Exception {
         }
     }
 
+    private static void runWithJUnitPlatform(Class<?> mainClass) throws Exception {
+        // https://junit.org/junit5/docs/current/user-guide/#launcher-api-execution
+        Thread.currentThread().setContextClassLoader(mainClass.getClassLoader());
```
jonathan-gibbons, line 114: this line, to set the context class loader, might be an issue for some core-libs tests.

src/share/classes/com/sun/javatest/regtest/agent/JUnitRunner.java
```diff
@@ -91,4 +109,40 @@ public static void main(ClassLoader loader, String... args) throws Exception {
...
+                StringWriter sw = new StringWriter();
+                try (PrintWriter pw = new PrintWriter(sw)) {
+                    pw.println("JavaTest Message: JUnit Platform Failure(s): " + summary.getTotalFailureCount());
+                    pw.println();
+                    for (TestExecutionSummary.Failure failure : summary.getFailures()) {
+                        failure.getException().printStackTrace(pw);
+                    }
+                    summary.printTo(pw);
+                }
+                System.err.println(sw);
```
jonathan-gibbons, line 140: printing to `System.err` may be an issue; this is happening on the test-side, isn't it?

test/junitTrace/JUnitTrace.gmk
```diff
@@ -26,17 +26,17 @@
...
 $(BUILDTESTDIR)/JUnitTrace.ok: \
-	    $(JTREG_IMAGEDIR)/lib/jtreg.jar \
-	    $(JTREG_IMAGEDIR)/bin/jtreg
+		$(JTREG_IMAGEDIR)/lib/jtreg.jar \
+		$(JTREG_IMAGEDIR)/bin/jtreg
 	$(RM) $(@:%.ok=%) ; $(MKDIR) $(@:%.ok=%)
 	$(JTREG_IMAGEDIR)/bin/jtreg $(JTREG_OPTS) \
 		-w:$(@:%.ok=%)/work -r:$(@:%.ok=%)/report \
 		-jdk:$(JDKHOME) \
 		-agentvm \
```
jonathan-gibbons, line 35: It might be better to upgrade this test to run in both modes (othervm, agentvm(), although maybe that could be could separately

src/share/classes/com/sun/javatest/regtest/agent/JUnitRunner.java
jonathan-gibbons: ... I guess it might just mean we can't use JUnit Jupiter for some of the low-level core-libs tests

src/share/classes/com/sun/javatest/regtest/agent/JUnitRunner.java
jonathan-gibbons: ... I guess it might just mean we can't use JUnit Jupiter for some of the low-level core-libs tests

src/share/classes/com/sun/javatest/regtest/agent/JUnitRunner.java
sormuras: Changed to `JUnit JAR file(s)`.

src/share/classes/com/sun/javatest/regtest/agent/JUnitRunner.java
sormuras: Yes, either that or we find another way to make those core-libs tests and JUnit happy by agreeing upon a class loader.

src/share/classes/com/sun/javatest/regtest/agent/JUnitRunner.java
sormuras: It's doing the same as the JUnit 4 runner already did before. Thus, it shouldn't introduce new issues.

test/junitTrace/JUnitTrace.gmk
sormuras: I agree — should be done in separate PR to keep the changes of this PR focussed around making the JUnit Platform Launcher work.

sormuras:
> Dummy comment to trigger bots.

Did bots what bots had to do?

sormuras:
>> Dummy comment to trigger bots.
>
> Did bots what bots had to do?

Ah, now bot removed the `ready` label. Good bot!

image

test/junitTrace/JUnitTrace.gmk
sormuras: Done via ...

src/share/classes/com/sun/javatest/regtest/agent/JUnitRunner.java
jonathan-gibbons: If there is a way to avoid using the context class loader and agree on some other one, that might be worth investigating as a followup.

src/share/classes/com/sun/javatest/regtest/agent/JUnitRunner.java
```diff
@@ -69,6 +79,14 @@ public static void main(ClassLoader loader, String... args) throws Exception {
             cl = JUnitRunner.class.getClassLoader();
         }
         Class<?> mainClass = Class.forName(className, false, cl);
+        if (JUNIT_RUN_WITH_JUNIT_4) {
+            runWithJUnit4(mainClass);
+        } else {
+            runWithJUnitPlatform(mainClass);
+        }
+    }
+
```
jonathan-gibbons, line 88: It's good to have the fallback for a while, but in time (7.1, 7.2) we should remove it.

jonathan-gibbons (approved):

PR #97, opened 2022-07-28
7903217: jtreg could try killing descendants of stuck test, before timing out the test

gerard-ziemski:
This is an enhancement that aims to improve the robustness of the testing by attempting to quit any child processes (that are possibly stuck and are blocking the parent process from terminating) before timing out the target parent process.

Aborting a process will flush its stdout/stderr streams, which will hopefully get captured in the test's log and provide additional clues as to why a test was timing out.

This enhancement was locally tested with a handcrafted test that itself launched a child process that would get stuck on purpose and worked as intended.

Hopefully, this will help debug issues such as [JDK-8286345](https://bugs.openjdk.org/browse/JDK-8286345)

---------
### Progress

### Issue
 * [CODETOOLS-7903217](https://bugs.openjdk.org/browse/CODETOOLS-7903217): jtreg could try killing descendants of stuck test, before timing out the test

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg pull/97/head:pull/97` \
`$ git checkout pull/97`

`$ git checkout pull/97` \
`$ git pull https://git.openjdk.org/jtreg pull/97/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 97`

`$ git pr show -t 97`

</details>

<a href="https://git.openjdk.org/jtreg/pull/97.diff">https://git.openjdk.org/jtreg/pull/97.diff</a>

</details>

### DIFF
--- a/src/share/classes/com/sun/javatest/diff/SuperDiff.java
+++ b/src/share/classes/com/sun/javatest/diff/SuperDiff.java
@@ -282,8 +282,8 @@ class SuperDiff extends Diff {
             return infoTable.get(path);
         }
 
-        final Set<String> platforms = new TreeSet<>();
-        final Map<String, Info> infoTable = new HashMap<>();
+        final transient Set<String> platforms = new TreeSet<>();
+        final transient Map<String, Info> infoTable = new HashMap<>();
     }
 
     static class YearDay implements Comparable<YearDay> {
--- a/src/share/classes/com/sun/javatest/regtest/exec/ProcessCommand.java
+++ b/src/share/classes/com/sun/javatest/regtest/exec/ProcessCommand.java
@@ -30,9 +30,12 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.PrintWriter;
+import java.lang.ProcessHandle;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+import java.util.stream.Stream;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 
@@ -243,15 +246,41 @@ public class ProcessCommand
             InputStream processErr = process.getErrorStream();
 
             long start = System.currentTimeMillis();
+            Alarm pre_alarm = Alarm.NONE;
             Alarm alarm = Alarm.NONE;
             final CountDownLatch timeoutHandlerDone = new CountDownLatch(1);
+            final long PRE_ALARM_TIME_MS = 5000;
 
             if (timeout > 0) {
                 final Thread victim = Thread.currentThread();
-                alarm = Alarm.schedule(timeout, TimeUnit.MILLISECONDS, out, new Runnable() {
-                    public void run() {
-                        invokeTimeoutHandler(timeoutHandler, timeoutHandlerDone, process, victim);
+                pre_alarm = Alarm.schedule(timeout, TimeUnit.MILLISECONDS, out, new Runnable() {
+                  // at the timeout, as the last resort, we will exit all the (possibly hanging),
+                  // descendants of the process, in the hope that doing so will unstuck the process itself
+                  public void run() {
+                    long pid = getTimeoutHandler().getProcessId(process);
+                    Optional<ProcessHandle> oph = ProcessHandle.of(pid);
+                    if (oph.isPresent()) {
+                      ProcessHandle processHandle = oph.get();
+                      Stream<ProcessHandle> descendants = processHandle.descendants();
+                      long count = descendants.count();
+                      if (count > 0) {
+                        System.err.println("Detected a possibly stuck process (pid:"+pid+") with "+count+" descendants");
+                        System.err.println("Attempting to kill:");
+                        descendants = processHandle.descendants();
+                        descendants.forEachOrdered(child -> System.err.println(" child process with pid:"+child.pid()));
+                        descendants = processHandle.descendants();
+                        descendants.forEachOrdered(child -> child.destroy());
+                        descendants = processHandle.descendants();
+                        descendants.forEachOrdered(child -> child.destroyForcibly());
+                      }
                     }
+                  }
+                });
+                alarm = Alarm.schedule(timeout+PRE_ALARM_TIME_MS, TimeUnit.MILLISECONDS, out, new Runnable() {
+                  // if we reach here, then we tried everything and the only option now is to timeout
+                  public void run() {
+                    invokeTimeoutHandler(timeoutHandler, timeoutHandlerDone, process, victim);
+                  }
                 });
             }
 
@@ -271,12 +300,14 @@ public class ProcessCommand
                 errCopier.join();
                 int exitCode = process.waitFor();
 
+                pre_alarm.cancel();
                 // if the timeout hasn't fired, cancel it as quickly as possible
                 alarm.cancel();
 
                 return getStatus(exitCode, statusScanner.exitStatus());
 
             } catch (InterruptedException e) {
+                pre_alarm.cancel();
                 alarm.cancel();
                 return Status.error("Program `" + cmd.get(0) + "' interrupted");
             } finally {
--- a/src/share/classes/com/sun/javatest/regtest/exec/ProcessCommand.java
+++ b/src/share/classes/com/sun/javatest/regtest/exec/ProcessCommand.java
@@ -254,51 +254,33 @@ public class ProcessCommand
             if (timeout > 0) {
                 final Thread victim = Thread.currentThread();
                 pre_alarm = Alarm.schedule(timeout, TimeUnit.MILLISECONDS, out, new Runnable() {
-                    // at the timeout, as the last resort, we will exit all the (possibly hanging),
-                    // descendants of the process, in the hope that doing so will unstuck the process itself
-                    public void run() {
-                        long pid = getTimeoutHandler().getProcessId(process);
-                        Optional<ProcessHandle> oph = ProcessHandle.of(pid);
-                        if (oph.isPresent()) {
-                            ProcessHandle processHandle = oph.get();
-                            Stream<ProcessHandle> descendants = processHandle.descendants();
-                            long count = descendants.count();
-                            if (count > 0) {
-                                OutputStream processOut = process.getOutputStream();  // input stream to process
-                                if (processOut != null) {
-                                    try {
-                                        processOut.write(new String("Detected a possibly stuck process (pid:"+pid+") with "+count+" descendants").getBytes());
-                                        processOut.write(new String("Attempting to kill:").getBytes());
-                                    } catch (IOException e) {
-                                        // ignore
-                                    }
-                                }
-                                descendants = processHandle.descendants();
-                                if (processOut != null) {
-                                    List<Long> pids = new java.util.ArrayList<Long>();
-                                    descendants.forEachOrdered(child -> pids.add(Long.valueOf(child.pid())));
-                                    try {
-                                        for (Long l : pids) {
-                                            processOut.write(new String(" child process with pid:"+l.longValue()).getBytes());
-                                        }
-                                        processOut.flush();
-                                    } catch (IOException e) {
-                                        // ignore
-                                    }
-                                }
-                                descendants = processHandle.descendants();
-                                descendants.forEachOrdered(child -> child.destroy());
-                                descendants = processHandle.descendants();
-                                descendants.forEachOrdered(child -> child.destroyForcibly());
-                            }
-                        }
+                  // at the timeout, as the last resort, we will exit all the (possibly hanging),
+                  // descendants of the process, in the hope that doing so will unstuck the process itself
+                  public void run() {
+                    long pid = getTimeoutHandler().getProcessId(process);
+                    Optional<ProcessHandle> oph = ProcessHandle.of(pid);
+                    if (oph.isPresent()) {
+                      ProcessHandle processHandle = oph.get();
+                      Stream<ProcessHandle> descendants = processHandle.descendants();
+                      long count = descendants.count();
+                      if (count > 0) {
+                        err.println("Detected a possibly stuck process (pid:"+pid+") with "+count+" descendants");
+                        err.println("Attempting to kill:");
+                        descendants = processHandle.descendants();
+                        descendants.forEachOrdered(child -> err.println(" child process with pid:"+child.pid()));
+                        descendants = processHandle.descendants();
+                        descendants.forEachOrdered(child -> child.destroy());
+                        descendants = processHandle.descendants();
+                        descendants.forEachOrdered(child -> child.destroyForcibly());
+                      }
                     }
+                  }
                 });
                 alarm = Alarm.schedule(timeout+PRE_ALARM_TIME_MS, TimeUnit.MILLISECONDS, out, new Runnable() {
-                    // if we reach here, then we tried everything and the only option now is to timeout
-                    public void run() {
-                        invokeTimeoutHandler(timeoutHandler, timeoutHandlerDone, process, victim);
-                    }
+                  // if we reach here, then we tried everything and the only option now is to timeout
+                  public void run() {
+                    invokeTimeoutHandler(timeoutHandler, timeoutHandlerDone, process, victim);
+                  }
                 });
             }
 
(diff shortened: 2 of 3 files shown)


src/share/classes/com/sun/javatest/diff/SuperDiff.java
```diff
@@ -282,8 +282,8 @@ Info getInfo(String path) {
             return infoTable.get(path);
         }
 
-        final Set<String> platforms = new TreeSet<>();
-        final Map<String, Info> infoTable = new HashMap<>();
+        final transient Set<String> platforms = new TreeSet<>();
+        final transient Map<String, Info> infoTable = new HashMap<>();
```
jonathan-gibbons, line 286: This seems totally unrelated

src/share/classes/com/sun/javatest/regtest/exec/ProcessCommand.java
```diff
@@ -271,12 +300,14 @@ public void run() {
                 errCopier.join();
                 int exitCode = process.waitFor();
 
+                pre_alarm.cancel();
                 // if the timeout hasn't fired, cancel it as quickly as possible
                 alarm.cancel();
 
                 return getStatus(exitCode, statusScanner.exitStatus());
```
jonathan-gibbons, line 307: I think you need to do something more here.   If the class killed child processes, you need to ensure that (probably) `Status.error` is returned with a suitable string. In other words, consider the case where you unlock a child process and the main process exits normally for some reason. That should *not* be grounds for the test to pass.

src/share/classes/com/sun/javatest/regtest/exec/ProcessCommand.java
```diff
@@ -243,15 +246,41 @@ public Status exec() {
...
+                  // descendants of the process, in the hope that doing so will unstuck the process itself
+                  public void run() {
+                    long pid = getTimeoutHandler().getProcessId(process);
+                    Optional<ProcessHandle> oph = ProcessHandle.of(pid);
+                    if (oph.isPresent()) {
+                      ProcessHandle processHandle = oph.get();
+                      Stream<ProcessHandle> descendants = processHandle.descendants();
+                      long count = descendants.count();
+                      if (count > 0) {
+                        System.err.println("Detected a possibly stuck process (pid:"+pid+") with "+count+" descendants");
```
jonathan-gibbons, line 267: Don't use `System.err` directly;  use the `err` and `out` streams that were passed in when the class was invoked.

src/share/classes/com/sun/javatest/diff/SuperDiff.java
gerard-ziemski: I wasn't able to compile jtreg without this modification, and I thought that including this change would help others in similar situation, but I can take it out and either leave it out completely or do a separate PR if you wish?

src/share/classes/com/sun/javatest/regtest/exec/ProcessCommand.java
gerard-ziemski: Thank you, will fix.

src/share/classes/com/sun/javatest/regtest/exec/ProcessCommand.java
gerard-ziemski: I see the point, but I also had a case where after force quitting the child process, the parent test process is able to retrieve everything it needed from the child (which finished its task required for the parent before getting itself locked up) and be able to succeed just fine.

If we were to return error code in such cases, we would be generating failures in cases where the parent process was able to finish, but its children were not. Not necessarily always directly related to the case under test, more of a tangential failure, but I do see the value in reporting such cases. I will tweak it then.

gerard-ziemski:
Thank you Jonathan for you feedback!

src/share/classes/com/sun/javatest/regtest/exec/ProcessCommand.java
gerard-ziemski: That got much more complicated that I imaged. Please take a look and tell me if I missed something and it could have been done in a simpler way.

gerard-ziemski: Hmm, I think I did overcomplicate things. After my last commit, I started working on the other feedback, and noticed "err" print writer, so I am using it now instead.

src/share/classes/com/sun/javatest/regtest/exec/ProcessCommand.java
gerard-ziemski: We now detect whether we had to cancel any child processes and return an error code if we do (this would fix any potential false positives, where the parent test process is free to proceed, but a child process is unable to throw any error status back at the parent, because the child itself is stuck)

gerard-ziemski:
Ping.

gerard-ziemski:
@jonathan-gibbons

hi Jonathan, is there anything I can do to help move this forward?

With this fix in, I think I will be able to make progress with issues such as [JDK-8286345](https://bugs.openjdk.org/browse/JDK-8286345)

src/share/classes/com/sun/javatest/diff/SuperDiff.java
jonathan-gibbons: It may be an artifact of the JDK you use to build jtreg.   I recommend using JDK 11.

jonathan-gibbons (commented):
I don't think the changes belong in `ProcessCommand`.

If they belong anywhere, they should be in timeout handling code.

gerard-ziemski:
> I don't think the changes belong in `ProcessCommand`.
> 
> If they belong anywhere, they should be in timeout handling code.

Do you want the timeout code to be able to change/influence this new (subprocess cleanup) phase? If so, how? If not, shouldn't we hide the details of this new step from the actual timeout handling code?

tstuefe:
Just a by-drive, I would prefer this to be opt-in and by default disabled. I like hanging processes to keep hanging in my tests, that way I can debug them the next morning if they are still around.

tstuefe:
Also, note that killing whole process trees in a quasi random order exposes you to the risk of zombification if you happen to kill branch processes before leafs. On a well-setup box usually no problem, since whoever adopts (pid 1, systemd, ...) periodically reaps. But it can be a problem in some container environments that don't set up a reaper (looking at you, Jenkins).

That is another reason to make this an optional feature.

gerard-ziemski:
> _Mailing list message from [Jonathan Gibbons](mailto:jonathan.gibbons@oracle.com) on [jtreg-dev](mailto:jtreg-dev@mail.openjdk.org):_
> 
> On 8/22/22 10:21 AM, Gerard Ziemski wrote:
> 
> > I see the point, but I also had a case where after force quitting the child process, the parent test process is able to retrieve everything it needed from the child (which finished its task required for the parent before getting itself locked up) and be able to succeed just fine.
> > If we were to return error code in such cases, we would be generating failures in cases where the parent process was able to finish, but its children were not. Not necessarily always directly related to the case under test, more of a tangential failure, but I do see the value in reporting such cases. I will tweak it then.
> 
> This sort of detailed analysis is beyond what is reasonable for `jtreg` itself.? When you want complex decision logic for "did it pass", you need to bake that into the test code.
> 
> -- Jon

I was trying to help debug a family of issues with the same problem - timeouts with missing timeout info, in hopes that killing/flushing subprocesses of a test that is about to fail (by timeout) anyhow might get us some extra debug info that could help.

It's an issue that affects a wide area of tests from all different components. For example a search of semi recent (2020-2022) bugs with:

```
Timeout information: 
--- Timeout information end. 
```
in common output, gave me this list of P1-P3 (I spent 10 minutes searching, but there are many many more):

https://bugs.openjdk.org/browse/JDK-8184445 JShell tests: fail intermittently if tests are run in high concurrent mode.
https://bugs.openjdk.org/browse/JDK-8286554 gc/stress/TestStressG1Humongous.java timed out
https://bugs.openjdk.org/browse/JDK-8288279 gc/z/TestHighUsage.java timed out
https://bugs.openjdk.org/browse/JDK-8251969 java/lang/invoke/RicochetTest.java timed out
https://bugs.openjdk.org/browse/JDK-8293289 gc/cslocker/TestCSLocker.java timed out
https://bugs.openjdk.org/browse/JDK-8270799 vmTestbase/nsk/jvmti/ tests timing out with JFR
https://bugs.openjdk.org/browse/JDK-8268379 java/util/Locale/LocaleProvidersRun.java and sun/util/locale/provider/CalendarDataRegression.java timed out
https://bugs.openjdk.org/browse/JDK-8265037 serviceability/sa/ClhsdbPmap.java#id1 failed with "RuntimeException: Process is still alive. Can't get its output."
https://bugs.openjdk.org/browse/JDK-8289918 serviceability/attach/AttachWithStalePidFile.java timed out with "IOException: Premature EOF"
https://bugs.openjdk.org/browse/JDK-8278369 java/nio/channels/Channels/TransferTo.java hangs in testStreamContents
https://bugs.openjdk.org/browse/JDK-8258648 vmTestbase/vm/mlvm/indy/stress/jdi/breakpointInCompiledCode/Test.java timed out
https://bugs.openjdk.org/browse/JDK-8249684 java/foreign/TestMismatch.java timed out

I was hoping that this enhancement would help with such issues to get child processes to flush their output to help debug them further.

It's not just one test, it's many.

tstuefe:
Hi Gerard,

What you are trying to do is useful appreciated. But I'm unsure too if handling this at jtreg level is the best thing. But I see two sides here and therefore keep out of the discussion.

But another thing, in order for this to be useful, we would need thread dumps from hanging children too, if the children happen to be JVMs. Just hoping that abort(3) will nudge the children enough to vomit some output will not often work. E.g. if jcmd hangs, it is usually innocent: it waits for an answer from the attachee, and that one is stuck. It would be perfectly able to react to a thread dump and tell me as much.

So, before killing them, sending each of them a SIGQUIT to get thread dumps and giving them a bit of time to respond. And that raises more questions. If you do this, especially wholesale for all children, you could absolutely flood the jtr files with thread dumps from children, and analyzing them gets really confusing.

Not sure what a good solution could be. Lets see what others think.

Cheers, Thomas

tstuefe:
Hi Gerard,

> > Hi Gerard,
> > What you are trying to do is useful and appreciated. I often was missing more info. But I'm unsure too if handling this at jtreg level is the best thing. But I see two sides here and therefore keep out of the discussion.
> > But another thing, in order for this to be useful, we would need thread dumps from hanging children too, if the children happen to be JVMs. Just hoping that abort(3) will nudge the children enough to vomit some output will not often work. E.g. if jcmd hangs, it is usually innocent: it waits for an answer from the attachee, and that one is stuck. It would be perfectly able to react to a thread dump and tell me as much.
> > So, before killing them, send each of them a SIGQUIT to get thread dumps and give them a bit of time to respond. And that raises more questions. If you do this, especially wholesale for all children, you could absolutely flood the jtr files with thread dumps from children, and analyzing them gets really confusing.
> > Not sure what a good solution could be. Let's see what others think.
> > Cheers, Thomas
> 
> Thank you Thomas for your feedback.
> 
> I really like your idea to SIGQUIT the children processes before force quitting them. And we can kill those intelligently to try and avoid leaving behind zombies.
> 
> There is no progress on many (all?) of the related bugs that I listed earlier. My hope is that logging more info from all processes involved would lead to a breakthrough that would allow the analysis to unblock. More info is the key in those issues I believe, so if there is an increase in some noise, then I think it is a fair price to pay.
> 
> Not sure what you mean by "flooding the jtr" issue - do you mean that too long jtr log files get trimmed ?

Yes, but also them being just plain indecipherable.

> I really dislike when that happens. Unsure when we decided to do this, but nowadays disk space should be plentiful to handle full log files without cutting them down, I would hope.


I'm a bit skeptical about finding one error that explains all. You may have to dive down into each one and do a targeted analysis.

wrt [JDK-8286345](https://bugs.openjdk.org/browse/JDK-8286345), I would be curious to see log files. The core was from jcmd or from the main program? I also would add some more checks to the OutputAnalyzer for the jcmd process (should have exited normally, etc) and maybe do a diagnostic report via output.reportDiagnosticSummary();

Note that ThreadedMallocTestType.java also relies on there being no other allocations with the mtTest category. That should be true, but is still a bit fragile.

Cheers, Thomas

gerard-ziemski:
> _Mailing list message from [Jonathan Gibbons](mailto:jonathan.gibbons@oracle.com) on [jtreg-dev](mailto:jtreg-dev@mail.openjdk.org):_
> 
> On 11/17/22 12:50 PM, Gerard Ziemski wrote:
> 
> If it's not one test but many, that suggests putting the functionality in code in a test-library, that can be reused by any necessary tests.
> 
> Also, note the possibility of using `@run driver` which is intended as an extension mechanism for customized execution models, such as that you are describing. In conjunction with test library code, it allows you to have complex execution models, which can even be different for different kinds of tests.?? You described one particular model where a process might time out but the test should still be deemed to have passed;? that may be reasonable for the tests you have in mind, but it does not sound general enough or standard enough to be baked into mainline jtreg.
> 
> If you have trouble modelling the behavior you want in a test driver class, then that is a reason to come back here and propose or ask for any necessary enhancements.

The goal here was to "flush" any possible outstanding output, not to make the test finish (either as pass or fail).

And in order to flush any outstanding output, we need to quit any lingering child processes. The other consequence of doing that is that the main test might now be able to continue and possibly finish. In that scenario, you have earlier raised a question on whether we should consider that pass or fail. I said originally, that it should pass (assuming the test's asserts all pass). However, the pass/fail issue was never at the issue here.

I really just wanted to unblock the error/output pipelines here of hanging processes that were about to timeout. So the effects of this fix would only apply to those tests that were about to time out anyhow.

I also was trying to make a point that this timeout, where we are missing possible outputs, affects random tests in a variety of components, so it's not possible to anticipate which ones would need this new feature, and I was trying to argue that we need it system wide, and I still think that.

We can introduce a switch that would make this optional, but that switch should affect ALL the tests.

On my part I don't understand why you both think that such approach should not be part of jtreg. The timeout mechanism itself is part of jtreg. I think it makes a perfect sense to extend it to terminate any hanging child processes (jtreg has a knowledge and a built-in mechanism for doing that, which would need to be re-implemented in any client using jtreg)

Can we continue discussing this?

On my part I will look into `@run driver` feature...

tstuefe:
> > _Mailing list message from [Jonathan Gibbons](mailto:jonathan.gibbons@oracle.com) on [jtreg-dev](mailto:jtreg-dev@mail.openjdk.org):_
> > On 11/17/22 12:50 PM, Gerard Ziemski wrote:
> > If it's not one test but many, that suggests putting the functionality in code in a test-library, that can be reused by any necessary tests.
> > Also, note the possibility of using `@run driver` which is intended as an extension mechanism for customized execution models, such as that you are describing. In conjunction with test library code, it allows you to have complex execution models, which can even be different for different kinds of tests.?? You described one particular model where a process might time out but the test should still be deemed to have passed;? that may be reasonable for the tests you have in mind, but it does not sound general enough or standard enough to be baked into mainline jtreg.
> > If you have trouble modelling the behavior you want in a test driver class, then that is a reason to come back here and propose or ask for any necessary enhancements.
> 
> The goal here was to "flush" any possible outstanding output, not to make the test finish (either as pass or fail).
> 
> And in order to flush any outstanding output, we need to quit any lingering child processes. The other consequence of doing that is that the main test might now be able to continue and possibly finish. In that scenario, you have earlier raised a question on whether we should consider that pass or fail. I said originally, that it should pass (assuming the test's asserts all pass). However, the pass/fail issue was never at the issue here.
> 
> I really just wanted to unblock the error/output pipelines here of hanging processes that were about to timeout. So the effects of this fix would only apply to those tests that were about to time out anyhow.
> 
> I also was trying to make a point that this timeout, where we are missing possible outputs, affects random tests in a variety of components, so it's not possible to anticipate which ones would need this new feature, and I was trying to argue that we need it system wide, and I still think that.
> 
> We can introduce a switch that would make this optional, but that switch should affect ALL the tests.
> 
> On my part I don't understand why you both think that such approach should not be part of jtreg. The timeout mechanism itself is part of jtreg. I think it makes a perfect sense to extend it to terminate any hanging child processes (jtreg has a knowledge and a built-in mechanism for doing that, which would need to be re-implemented in any client using jtreg)
> 
> Can we continue discussing this?
> 
> On my part I will look into `@run driver` feature...

Hi Gerard,

sorry for the delay.

I am fine with adding it to jtreg as an optional feature. I don't know enough about jtreg to review this in detail though.

What would be nice is if you were to send SIGQUIT to the processes which are java processes, to have a chance to get thread dumps.

Cheers, Thomas

gerard-ziemski:
I'm going to withdraw this issue and see if I can do this with a custom timeout handler mechanism.

PR #115, opened 2022-09-09
CODETOOLS-7903267: Support executing a single test method in a JUnit class

jonathan-gibbons:
Please review a medium-sized patch to support new feature that provides a way to specify that only parts of a test should be executed.  For JUnit tests, this allows a single method in the test to be executed, but the underlying mechanism is more general than that and can be leveraged by any framework that supports running many test cases within an overall test.

The feature is enabled by using a new syntactic form on the command line, informally called _query syntax_: `path-to-test?string`. The syntax is based on URL syntax, but note that unlike URLs, the query component comes at the end, and not before any fragment component. In other words, the full syntax is `path#id?string`. Generally, this reflects the hierarchical structure:   a path to a file, a test description within the file, and a string to specify part of the test to be run.

In terms of implementation, the "heavy lifting" is done in `TestManager`, to track the _file, id, query_ tuple, as compared to the previous _file, id_.  New abstractions `TestSpec` and `GroupSpec` are introduced, to replace stylistic use of simple strings.

The underlying JT Harness does not know (or need to know) about query components and so is unchanged. But the back end `RegressionScript` used to execute each test _does_ need to know the query component, and so the information about which tests involve a query component is tunneled through the `RegressionParameters` in a side channel, embodied by a new resource in the object. In itself, the `RegressionScript` has little to do with the query component, it is simply passed down to the actions of the test in a new `test.query` property or `TESTQUERY` environment variable.

`JUnitRunner` is updated to check for the new `test.query` property: if set, it is used as the name of a test method to be run. Note: a JUnit exception will be thrown and the test will fail if the method is not found. The exception does contain enough useful information to diagnose the issue, but we may want to detect and report the issue in a more friendly form in a followup change.

Finally, new jtreg self-tests are added to exercise the new feature.

---------
### Progress

### Issue
 * [CODETOOLS-7903267](https://bugs.openjdk.org/browse/CODETOOLS-7903267): Support executing a single test method in a JUnit class

### Reviewers
 * [Iris Clark](https://openjdk.org/census#iris) (@irisclark - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg pull/115/head:pull/115` \
`$ git checkout pull/115`

`$ git checkout pull/115` \
`$ git pull https://git.openjdk.org/jtreg pull/115/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 115`

`$ git pr show -t 115`

</details>

<a href="https://git.openjdk.org/jtreg/pull/115.diff">https://git.openjdk.org/jtreg/pull/115.diff</a>

</details>

### DIFF
--- a/src/share/classes/com/sun/javatest/regtest/config/TestManager.java
+++ b/src/share/classes/com/sun/javatest/regtest/config/TestManager.java
@@ -217,7 +217,7 @@ public class TestManager {
          * The form contains the file and id as a relative URL.
          * It does not include the query.
          *
-         * @return
+         * @return the path for a test
          */
         String getTestPath() {
             return id == null ? pathToString(file) : pathToString(file) + "#" + id;
--- a/src/share/classes/com/sun/javatest/regtest/config/TestManager.java
+++ b/src/share/classes/com/sun/javatest/regtest/config/TestManager.java
@@ -38,12 +38,16 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
+import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Properties;
 import java.util.Set;
 import java.util.TreeMap;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
-import com.sun.javatest.TestFilter;
 import com.sun.javatest.TestFinder;
 import com.sun.javatest.TestResult;
 import com.sun.javatest.TestResultTable;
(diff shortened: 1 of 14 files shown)


test/junitQueryTest/JUnitQueryTest.gmk
```diff
@@ -0,0 +1,246 @@
+#
+# Copyright (c) 2022, Oracle and/or its affiliates. All rights reserved.
+# DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+#
+# This code is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 only, as
+# published by the Free Software Foundation.  Oracle designates this
+# particular file as subject to the "Classpath" exception as provided
```
irisclark, line 8: The use of GPL+CPE seems unusual for a test.

test/junitQueryTest/JUnitQueryTest.gmk
```diff
@@ -0,0 +1,246 @@
...
+		-jdk:$(JDKHOME) \
+		-va \
+		$(TESTDIR)/junitQueryTest/a/b/c?m13 \
+			> $(@:%.ok=%)/jt.log 2>&1 || \
+		true "non-zero exit code from jtreg intentionally ignored"
+	$(GREP) -s "Error: Invalid use of query component: a/b/c?m13" $(@:%.ok=%)/jt.log
+	echo $@ passed at `date` > $@
+
+TESTS.jtreg += $(BUILDTESTDIR)/JUnitQueryTest.invalidQuery.ok
+
```
irisclark, line 233: I'm sure there are others, but behavior when the query is associated with a valid test springs to mind, i.e.

Negative test: a/b/c/Test1.java?m14 

Also test: a/b/c/Test1.java?m14 and a/b/c/Test.java

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
```diff
@@ -1125,9 +1101,11 @@ public void error(String msg) {
                 Tool.this.error(msg);
             }
         });
-        testManager.addTestFiles(testFileArgs);
-        testManager.addTestFileIds(testFileIdArgs);
-        testManager.addGroups(testGroupArgs);
+//        testManager.addTestFiles(testFileArgs);
```
irisclark, line 1104: Did you intend to retain these commented lines?

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
```diff
@@ -2230,9 +2217,8 @@ private boolean isCygwinDetected() {
     }
 
     /**
-     * Returns whether or not Windows Subsystem for Linux may be
-     * available, by examining to see whether wsl.exe can be found on
-     * the path.
+     * Returns whether Windows Subsystem for Linux may be available,
+     * by examining to see whether wsl.exe can be found on the path.
```
irisclark, line 2221: I think that "to see" is redundant.

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
```diff
@@ -2288,9 +2274,11 @@ private static File getNormalizedFile(File f) {
     private Float timeoutFactorArg;
     private String priorStatusValuesArg;
     private Path reportDirArg;
-    public List<String> testGroupArgs = new ArrayList<>();
-    public List<Path> testFileArgs = new ArrayList<>();
-    public List<TestManager.FileId> testFileIdArgs = new ArrayList<>();
+//    public List<String> testGroupArgs = new ArrayList<>();
```
irisclark, line 2277: Intentionally left?

irisclark (commented):
Fantastic feature to allow developers to conveniently run specific tests during development!

I appreciate the addition documentation (javadoc and comments).

irisclark (approved):

LudwikJaniuk:
Does the new syntax allow skipping that part of the test? I.e., can I use the new granularity for exclusion, as well as inclusion?

sormuras:
As the title of this issue and pull requests states, this new syntax only supports running one specific test (method) of a JUnit test (class).

A more general test selection and filter feature could be described and implemented in a separate feature request. Maybe supporting (a subset of) all JUnit Platform's selection and filter options: https://junit.org/junit5/docs/current/user-guide/#running-tests-console-launcher-options

src/share/classes/com/sun/javatest/regtest/config/TestManager.java
```diff
@@ -76,90 +80,283 @@ public NoTests() {
...
+            this.id = id;
+            this.query = query;
+        }
+
+        /**
+         * Returns the path for a test, as required by JavaTest.
+         * The form contains the file and id as a relative URL.
+         * It does not include the query.
+         *
+         * @return
```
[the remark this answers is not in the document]
jonathan-gibbons: No, will update or remove

src/share/doc/javatest/regtest/faq.md
```diff
@@ -482,6 +489,14 @@ Note that in addition to the command-line options just listed, a test
 may contain tags such as `@requires` and `@modules` that determine whether
 a test should be run on any particular system.
 
+### How do I specify to run a single test method in a JUnit test?
```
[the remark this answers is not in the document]
jonathan-gibbons: Thanks; will clean up

test/junitQueryTest/JUnitQueryTest.gmk
jonathan-gibbons: will fix

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
jonathan-gibbons: no, I thought I'd removed them all

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
jonathan-gibbons: 👍

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
jonathan-gibbons: no

jonathan-gibbons:
> As the title of this issue and pull request states, this new syntax only supports running one specific test (method) of a JUnit test (class).
> 
> A more general test selection and filter feature could be described and implemented in a separate feature request. Maybe supporting (a subset of) all JUnit Platform's selection and filter options: https://junit.org/junit5/docs/current/user-guide/#running-tests-console-launcher-options

Yes, but ...

The primary design center of the jtreg support for TestNG and JUnit to to bulk run these tests as part of a larger test run, perhaps as part of a test group or tier.  The ability to run a single method is potentially interesting, for debugging purposes. The more that folk want to use more of the TestNG and JUnit features, the more they should be using those tools directly, or perhaps using an IDE, especially for "self-contained" tests.  The exception is probably the tests that require features like native code libraries that is more conveniently provided by the "make test" feature in the makefiles.

jonathan-gibbons:
> As the title of this issue and pull request states, this new syntax only supports running one specific test (method) of a JUnit test (class).

While the motivation and code here is for JUnit, the mechanism is intended to be applicable to other test frameworks. There are a variety of such frameworks in the JDK test base, but we'll have to wait for this version of jtreg to be the default before we can use this feature in those frameworks.

src/share/doc/javatest/regtest/faq.md
```diff
@@ -434,12 +434,18 @@ the sections containing any of the specified terms. For example:
...
+
 If a file contains multiple tests, you can specify the name of a test within
 that file by appending `#`_ID_ to the file path, where _ID_ is either defined
 in the test itself after the `@test` tag, or the string `id`_N_  if no id is 
 explicitly defined, where N is the number of the test within the file, 
 and where `0` identifies the first test. 
 
+If you specify `?`_string_ after the name of a test, the _string_ will be
+passed down to the test, for the test to filter the parts of the test to be
+executed. For any tests executed by JUnit Platform, the string is interpreted
```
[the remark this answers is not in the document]
jonathan-gibbons: done

src/share/doc/javatest/regtest/faq.md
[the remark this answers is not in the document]
jonathan-gibbons: ✅

test/junitQueryTest/JUnitQueryTest.gmk
jonathan-gibbons: Actually, this is a `*.gmk` file, and the file is in line with the existing practice of using GPL+CE for all makefiles, including test makefiles.

test/junitQueryTest/JUnitQueryTest.gmk
jonathan-gibbons: I can add a negative test, such as for `?m14` but you just get an exception from JUnit about "method not found", so it counts as a "vanilla" test failure.

This would extend to the second proposed example: the tests in the file without a query would be executed, and the test with a bad query would fail.

Thinking about whether `jtreg` could detect and report an *error* instead of the test *failed*, I guess we _could_ check for the existence of the method in `JUnitAction` (i.e. within jtreg itself) but this would require that `jtreg` is both compiled and run with a JDK at least as new as the class being tested ... and that would be bad.

Once we are in the test runner or test code, the only options are to pass or fail -- i.e. return normally or throw an exception. We could carve out a special case similar to `jtreg.SkipException` but I'm not convinced it is worth the effort.   If the user gives a bad name, you get an exception.

```
TEST RESULT: Failed. Execution failed: `main' threw exception: org.junit.platform.commons.JUnitException: TestEngine with ID 'junit-jupiter' failed to discover tests
```

```
org.junit.platform.commons.JUnitException: TestEngine with ID 'junit-jupiter' failed to discover tests
Caused by: org.junit.platform.commons.JUnitException: MethodSelector [className = 'Test1', methodName = 'm14', methodParameterTypes = ''] resolution failed
	... 15 more
Caused by: org.junit.platform.commons.PreconditionViolationException: Could not find method with name [m14] in class [Test1].
	... 20 more
```

jonathan-gibbons:
> Does the new syntax allow skipping that part of the test? I.e., can I use the new granularity for exclusion, as well as inclusion?

No, not at this time.  It would be better to debug and fix the test method so that it does not need to be excluded!

However, as far as jtreg is concerned, there is no semantics to the query string; it is up to the test (runner) to interpret the string, and that could support an extended syntax, such as exclusion `test-path?-method` or a list `test-path?m1,m2`.
But I think we should wait to see a compelling use case for such forms.  

Note also that you can always use JUnit Platform directly to execute a test, and with that, you have the full power of the JUnit platform command line.

irisclark (approved):
Updates look good.  Thanks!

PR #117, opened 2022-09-11
7903300: Add test library support to jtreg plugin

lgxbslgx:
Hi all,

Currently, when we open a jtreg test file which uses the JUnit or TestNG, the plugin can't identify the related classes of the JUnit or TestNG. So the statement such as `import org.junit.Test` is marked as red with error message `Cannot resolve symbol 'junit'`.

This patch adds the jtreg libraries directory to the lib path of the `Jetbrains IDEA` when opening the JUnit or TestNG test files so that the `Jetbrains IDEA` can identify the related classes.

Thanks for taking the time to review.

Best Regards,
-- Guoxiong

---------
### Progress

### Issue
 * [CODETOOLS-7903300](https://bugs.openjdk.org/browse/CODETOOLS-7903300): Add test library support to jtreg plugin (**Enhancement** - P4)

### Reviewers
 * [Maurizio Cimadamore](https://openjdk.org/census#mcimadamore) (@mcimadamore - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/117/head:pull/117` \
`$ git checkout pull/117`

`$ git checkout pull/117` \
`$ git pull https://git.openjdk.org/jtreg.git pull/117/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 117`

`$ git pr show -t 117`

</details>

<a href="https://git.openjdk.org/jtreg/pull/117.diff">https://git.openjdk.org/jtreg/pull/117.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/117#issuecomment-1242961941)

### DIFF
--- a/plugins/idea/src/main/java/com/oracle/plugin/jtreg/components/JTRegFileManagerListener.java
+++ b/plugins/idea/src/main/java/com/oracle/plugin/jtreg/components/JTRegFileManagerListener.java
@@ -141,8 +141,8 @@ public class JTRegFileManagerListener implements FileEditorManagerListener {
                     }
                 }
                 if (isTestNg || isJUnit) {
-                    String dir = "file://" + JTRegService.getInstance(project).getJTRegDir() + File.separator + "lib";
-                    Library library = JTRegUtils.createJTRegLibrary(project, dir, dir);
+                    String libDir = JTRegService.getInstance(project).getJTRegDir();
+                    Library library = JTRegUtils.createJTRegLibrary(project, libDir);
                     testInfo.jtregLib = library;
                     rootModel.addLibrary(library);
                 } else if (testInfo.jtregLib != null) {
--- a/plugins/idea/src/main/java/com/oracle/plugin/jtreg/components/JTRegFileManagerListener.java
+++ b/plugins/idea/src/main/java/com/oracle/plugin/jtreg/components/JTRegFileManagerListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2017, 2019 Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2017, 2022, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -39,9 +39,12 @@ import com.intellij.openapi.module.ModuleUtilCore;
 import com.intellij.openapi.project.DumbService;
 import com.intellij.openapi.project.Project;
 import com.intellij.openapi.roots.*;
+import com.intellij.openapi.roots.libraries.Library;
 import com.intellij.openapi.vfs.VirtualFile;
 import com.intellij.util.Alarm;
+import com.oracle.plugin.jtreg.service.JTRegService;
 import com.oracle.plugin.jtreg.util.JTRegUtils;
+import java.io.File;
 import org.jetbrains.annotations.NotNull;
 
 import java.util.*;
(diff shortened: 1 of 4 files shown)


plugins/idea/src/main/java/com/oracle/plugin/jtreg/components/JTRegFileManagerListener.java
```diff
@@ -134,6 +140,15 @@ void processFileOpened(TestInfo testInfo) {
                         rootModel.addSourceFolders(testRoots);
                     }
                 }
+                if (isTestNg || isJUnit) {
+                    String dir = "file://" + JTRegService.getInstance(project).getJTRegDir() + File.separator + "lib";
+                    Library library = JTRegUtils.createJTRegLibrary(project, dir, dir);
+                    testInfo.jtregLib = library;
+                    rootModel.addLibrary(library);
+                } else if (testInfo.jtregLib != null) {
+                    rootModel.removeLibrary(testInfo.jtregLib);
```
mcimadamore, line 149: I'm not sure this is logically correct. While I can understand adding jtreg/testng library to the model as soon as a new test file requiring it is opened, removing the library seems more problematic. In principle we should remove the testng/junit library from the model only if there's no open test that's a jtreg test. Which doesn't seem to be what's happening in the code.

mcimadamore:
Note that there is a workaround for the problem mentioned in this PR (not that I think it wouldn't be nice to fix). The IDE will typically suggest to add relevant test libraries when clicking on the light-bulb fix icon beside a TestNG/JUnit import.

plugins/idea/src/main/java/com/oracle/plugin/jtreg/util/JTRegUtils.java
```diff
@@ -315,6 +320,52 @@ public static boolean isTestNGImport(PsiImportStatement importStatement) {
...
+     */
+    public static boolean isJunitImport(PsiImportStatement importStatement) {
+        String qualifiedName = importStatement.getQualifiedName();
+        return qualifiedName != null && qualifiedName.startsWith("org.junit");
+    }
+
+    /**
+     * Create (if not existing) and get the jtreg project library.
+     */
+    public static Library createJTRegLibrary(Project project, String oldDir, String newDir) {
```
mcimadamore, line 352: Wouldn't it be better to split this into a function that creates the library, and one that updates it? I find it odd that the file listener is calling this with oldDir == newDir. Maybe just add option for oldDir to be null, so that we only drop the old root if one is given?

mcimadamore:
Also, when updating the plugin, it is better to update the `gradle.properties` file to bump the version, and to include some notes (e.g. bug number and synopsis). This will allow upgrade to occur smoothly (and issues to be pinpointed correctly).

plugins/idea/src/main/java/com/oracle/plugin/jtreg/util/JTRegUtils.java
lgxbslgx: I adjusted the api, added `update` method and delegated the `create` method to it. And I also fixed the possible NPE and avoided the duplicated lib entries.

lgxbslgx:
> Also, when updating the plugin, it is better to update the `gradle.properties` file to bump the version, and to include some notes (e.g. bug number and synopsis). This will allow upgrade to occur smoothly (and issues to be pinpointed correctly).

I bumped the version and revised the notes.

I think the plugin should be released periodically (may be several months). But now it seems you want to release a new version one patch. I don't know the current release flow so I follow your command.

mcimadamore (approved):
Looks good

lgxbslgx:
@mcimadamore Thanks for the review. Still waiting for a formal reviewer to review it.

lgxbslgx:
Thanks for the review. Could I get your help to sponsor this patch?

/integrate

PR #136, opened 2022-10-21
7903373: Add ability to customize test execution using main wrapper plugin

lmesnik:
The fix adds support of a plugin that customizes test execution.  

---------
### Progress

### Issue
 * [CODETOOLS-7903373](https://bugs.openjdk.org/browse/CODETOOLS-7903373): Add ability to customize test execution using main wrapper plugin

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg pull/136/head:pull/136` \
`$ git checkout pull/136`

`$ git checkout pull/136` \
`$ git pull https://git.openjdk.org/jtreg pull/136/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 136`

`$ git pr show -t 136`

</details>

<a href="https://git.openjdk.org/jtreg/pull/136.diff">https://git.openjdk.org/jtreg/pull/136.diff</a>

</details>


src/share/classes/com/sun/javatest/regtest/tool/i18n.properties
```diff
@@ -171,6 +171,10 @@ help.main.ignore.error.desc=(Default.) Execute the actions up to the @ignore tag
     then give an "Error" result.
 help.main.ignore.run.desc=Run the test, as though the @ignore tag were not present.
 help.main.l.desc=List the tests that would be executed instead of executing them.
+help.main.mw.arg=Man wrapper argument
+help.main.mw.desc=Man wrapper desc
+help.main.mwp.arg=Man wrapper argument
+help.main.mwp.desc=Man wrapper desc
```
jonathan-gibbons, line 177: possible spelling errors: should "Man wrapper" be "Main" -- if not, what is a "Man Wrapper" ?

test/mainWrapper/TestMainWrapper.gmk
```diff
@@ -0,0 +1,49 @@
...
+	JTREG_JAVA=$(JDKJAVA) $(JTREG_IMAGEDIR)/bin/jtreg $(JTREG_OPTS) \
+		-jdk:$(JDKHOME) \
+                -agentvm \
+		-w:$(@:%.ok=%)/cmd.agentvm/work \
+		-r:$(@:%.ok=%)/cmd.agentvm/report \
+		-verbose:fail \
+		$(TESTDIR)/mainWrapper
+	echo "test passed at `date`" > $@
+
+INITIAL_TESTS += $(BUILDTESTDIR)/TestMainWrapper.ok
```
jonathan-gibbons, line 49: I don't think this test is important enough to be in the set of initial tests; it should just be in the set of general jtreg tests (`TESTS.jtreg`)

jonathan-gibbons (commented):
On a first pass, it generally looks very good;  well done.

I'd like to go through it again next week.

test/mainWrapper/TestMainWrapper.gmk
lmesnik: fixed

src/share/classes/com/sun/javatest/regtest/tool/i18n.properties
lmesnik: fixed

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
```diff
@@ -1203,6 +1219,17 @@ public int run() throws BadArgs, Fault, Harness.Fault, InterruptedException {
...
 
+        if (customMainWrapper != null) {
+            CustomMainWrapper cmw = CustomMainWrapper.getInstance(customMainWrapper, customMainWrapperPathArg);
+            testVMOpts.add("-D" + MainWrapper.MAIN_WRAPPER + "=" + customMainWrapper);
+            testVMOpts.addAll(cmw.getAdditionalVMOpts());
+        }
+
+        if (customMainWrapperPathArg != null) {
+            testVMOpts.add("-D" + MainWrapper.MAIN_WRAPPER_PATH + "=" + customMainWrapperPathArg);
+        }
```
jonathan-gibbons, line 1230: It seems premature to be stashing values in the `testVMOpts` here.  

The more conventional way would be to pass these values through the `RegressionParameters` object to where the instance of the `RegressionScript` can decide whether to take these values into account for each individual test.

src/share/classes/com/sun/javatest/regtest/agent/CustomMainWrapper.java
```diff
@@ -0,0 +1,105 @@
...
+    @Override
+    public Thread createThread(ThreadGroup tg, Runnable task) {
+        return new TestThread(tg, task);
+    }
+
+    @Override
+    public List<String> getAdditionalVMOpts() {
+        return vmOpts;
+    }
+}
```
jonathan-gibbons, line 105: Is this real code, or just toy test code?

jonathan-gibbons: It loos like it is just test code, included from the new test, in which case it doesn't belong here in the mainline source.

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
```diff
@@ -1203,6 +1219,17 @@ public int run() throws BadArgs, Fault, Harness.Fault, InterruptedException {
             reportDirArg = Path.of("JTreport");
         }
 
+        if (customMainWrapper != null) {
+            CustomMainWrapper cmw = CustomMainWrapper.getInstance(customMainWrapper, customMainWrapperPathArg);
+            testVMOpts.add("-D" + MainWrapper.MAIN_WRAPPER + "=" + customMainWrapper);
+            testVMOpts.addAll(cmw.getAdditionalVMOpts());
```
jonathan-gibbons, line 1225: Ah, OK, there's the possibility that the instance of the wrapper class might want to specify some VM options. That's somewhat circular.  https://en.wikipedia.org/wiki/Ouroboros

src/share/classes/com/sun/javatest/regtest/agent/CustomMainWrapper.java
```diff
@@ -0,0 +1,82 @@
...
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Interface which should be implemented to customize execution of test.
+ * It is used by main and driver actions to execute test.
```
jonathan-gibbons, line 38: This comment needs a lot of work. More than anywhere else, this seems like the place to document this feature.  In particular, the only aspect of the execution that can be customized is the thread that is created to run the test.

src/share/classes/com/sun/javatest/regtest/agent/CustomMainWrapper.java
```diff
@@ -0,0 +1,82 @@
...
+                } catch (MalformedURLException e) {
+                }
+            }
+            loader = new URLClassLoader(urls.toArray(new URL[urls.size()]), loader);
+        }
+        try {
+            Class<? extends CustomMainWrapper> clz = loader.loadClass(className).asSubclass(CustomMainWrapper.class);
+            Constructor<? extends CustomMainWrapper> ctor = clz.getDeclaredConstructor();
+            CustomMainWrapper wrapper = ctor.newInstance();
+            wrapper.setAction(actionName);
```
jonathan-gibbons, line 61: Is the `actionName` ever going to be useful?   Di you use it in the Loom implementation?  Why would the thread created by `createThread` depend on the difference between `main` and `driver` actions?

(In general, a `driver` action is handled much earlier in the execution, when `RegressionScript` decides what VM options (if any) should be used for the agent.

jonathan-gibbons (commented):
I think there should be more documentation for this feature, both internal and for end users.

Generally, I find the naming based on "main wrapper" and "custom main wrapper" to be confusing and/or misleading, although you might be able to work around that with better documentation.

At the core, the name `CustomMainWrapper` suggests a variant, perhaps a subtype, of `MainWrapper`, but that's not the case.`MainWrapper` delegates to `CustomMainWrapper`,
and (somewhat more weirdly) `MainActionHelper` uses `CustomMainWrapper` as well.

In fact, `CustomMainWrapper` doesn't seem to be a custom main-wrapper at all; it is really just a custom thread factory, as can be seen by the only uses in the two instances of the expression `CustomMainWrapper.getInstance(...).createThread()`.  Indeed, `createThread` is the only real abstract method on the interface; `setAction` is just there for use during initialization, since the instance of `CMW.getInstance` is discarded after a thread has been created.

In terms of documentation, I suggest:

* better, more accurate, more specific documentation on the new `CustomMainWrapper` class. In particular, this documentation should indicate that the class just provides the ability to specify a thread factory to be used by the standard `MainWrapper` class

* more info in the i18.properties file (which is used to generate the command-line help.)
 In particular, although I have not compiled and run the code, I suspect that the command-line help for the `-mainwrapper` option will look something like
       `-mw <classname>, -mainwrapper <classname>   Main wrapper classname.`
which is essentially content-free, giving no useful info at all to the reader

* in the tag spec, in the appropriate Appendix, you should list any system properties that are set during this feature.

src/share/classes/com/sun/javatest/regtest/agent/CustomMainWrapper.java
```diff
@@ -0,0 +1,82 @@
...
+            return wrapper;
+        } catch (Exception e) {
```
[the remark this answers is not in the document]
lmesnik: fixed

src/share/classes/com/sun/javatest/regtest/agent/CustomMainWrapper.java
```diff
@@ -0,0 +1,82 @@
...
+
+/**
+ * Interface which should be implemented to customize execution of test.
+ * It is used by main and driver actions to execute test.
+ */
+public interface CustomMainWrapper {
+    static CustomMainWrapper getInstance(String mainWrapper, String path) {
+        String[] args = mainWrapper.split(":", 2);
+        String className = args[0];
+        String actionName = args[1].split("=")[1];
```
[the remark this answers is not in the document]
lmesnik: The actionName is not used anymore.

src/share/classes/com/sun/javatest/regtest/agent/CustomMainWrapper.java
lmesnik: I re-checked, seems actionName might be not needed. We could use virtual threads in all actions. So I removed actionName.

src/share/classes/com/sun/javatest/regtest/agent/CustomMainWrapper.java
```diff
@@ -0,0 +1,82 @@
...
+            wrapper.setAction(actionName);
+            return wrapper;
+        } catch (Exception e) {
+            throw new RuntimeException(e);
+        }
+    }
+
+    /**
+     * This method should be implemented to run test task.
+     * Default jtreg implementation is return new Thread(tg, task);
```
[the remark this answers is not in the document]
lmesnik: I removed this line. The overall CustomMainWrapper documentation should be enough.

src/share/classes/com/sun/javatest/regtest/agent/CustomMainWrapper.java
```diff
@@ -0,0 +1,82 @@
...
+     * @param tg ThreadGroup to run test
+     * @param task The task which executes test
+     */
+    Thread createThread(ThreadGroup tg, Runnable task);
+
+    /**
+     * This method is used to get information about current action.
+     * @param actionName name of action
+     */
+    default void setAction(String actionName) {
```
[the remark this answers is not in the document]
lmesnik: The method is removed.

src/share/classes/com/sun/javatest/regtest/config/RegressionParameters.java
```diff
@@ -629,6 +629,11 @@ public boolean getCacheableValue(TestDescription td) throws Fault {
     private static final String TIMEOUT_HANDLER = ".timeoutHandler";
     private static final String TIMEOUT_HANDLER_PATH = ".timeoutHandlerPath";
     private static final String TIMEOUT_HANDLER_TIMEOUT = ".timeoutHandlerTimeout";
+
+    private static final String CUSTOM_MAIN_WRAPPER = ".customMainWrapper";
+
```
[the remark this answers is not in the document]
lmesnik: Removed.

src/share/classes/com/sun/javatest/regtest/exec/MainAction.java
```diff
@@ -620,8 +627,11 @@ private Status runAgentJVM() throws TestRunException {
 
         Agent agent;
         try {
+            String wrapper = script.getCustomWrapper() == null ? null : script.getCustomWrapper() + ":" + "action=" + this.getName();
```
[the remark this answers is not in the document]
lmesnik: the action is removed

src/share/classes/com/sun/javatest/regtest/agent/CustomMainWrapper.java
lmesnik: Comments were updated.

src/share/classes/com/sun/javatest/regtest/exec/MainAction.java
```diff
@@ -480,6 +480,13 @@ private Status runOtherJVM() throws TestRunException {
         else if (policyFN != null) {
             javaProps.put("java.security.manager", "default");
         }
+
+        if (script.getCustomWrapper() != null) {
+            javaProps.put(MainWrapper.MAIN_WRAPPER, script.getCustomWrapper() + ":" + "action=" + this.getName());
+        }
+        if (script.getCustomWrapperPath() != null) {
+            javaProps.put(MainWrapper.MAIN_WRAPPER_PATH, script.getCustomWrapperPath());
+        }
```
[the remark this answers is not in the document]
lmesnik: These properties are internal, seems such properties are not documented in tag spec.

src/share/classes/com/sun/javatest/regtest/agent/TestThreadFactory.java
```diff
@@ -0,0 +1,84 @@
...
+import java.net.MalformedURLException;
+import java.net.URL;
+import java.net.URLClassLoader;
+import java.nio.file.Path;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * {@code TestThreadFactory} allows some customization of test execution.
+ * The jtreg creates new thread for each test using {@code new Thread(ThreadGroup tg, Runnable task);}.
```
jonathan-gibbons, line 39: Recommend: ~The~ `jtreg creates a new thread`

src/share/classes/com/sun/javatest/regtest/tool/i18n.properties
```diff
@@ -171,6 +171,12 @@ help.main.ignore.error.desc=(Default.) Execute the actions up to the @ignore tag
     then give an "Error" result.
 help.main.ignore.run.desc=Run the test, as though the @ignore tag were not present.
 help.main.l.desc=List the tests that would be executed instead of executing them.
+help.main.ttf.arg=<classname>
+help.main.ttf.desc=Specifies the class to create thread factory for test execution. \
+    The class should implement spcifig interface.
```
jonathan-gibbons, line 176: Typo: `spcifig` ?

src/share/classes/com/sun/javatest/regtest/tool/i18n.properties
```diff
@@ -171,6 +171,12 @@ help.main.ignore.error.desc=(Default.) Execute the actions up to the @ignore tag
...
+help.main.ttfp.arg=<path>
+help.main.ttfp.desc=Specifies classspath with custom test thread factory implementation. \
```
jonathan-gibbons, line 178: Suggest: replace `with` by `for`

test/docs.gmk
```diff
@@ -29,6 +29,5 @@ $(BUILDTESTDIR)/tidy-docs.ok: \
 	$(TIDY) -e --gnu-emacs true -q $<
 	echo "docs checked at `date`" > $@
 
-INITIAL_TESTS += \
-	$(BUILDTESTDIR)/tidy-docs.ok
-
+#INITIAL_TESTS += \
+#	$(BUILDTESTDIR)/tidy-docs.ok
```
jonathan-gibbons, line 33: Why is this commented out?  That seems unrelated.

src/share/classes/com/sun/javatest/regtest/agent/TestThreadFactory.java
jonathan-gibbons: The name is substantially better than before (it used to be `CustomMainWrapper`).

Now, the name is annoyingly ambiguous.   It can be read as `Test` (imperative verb) `Thread-Factory` (noun) or `Test-Thread` (adjective) `Factory` (noun). (I believe it is the latter that is intended.)

I can't think of a better name, and I don't want to suggest we should go through another name change again!
For now, I think it may be enough to add a new explanatory sentence at the beginning of the doc comment,
and/or work on the rest of the comment text.
```
A factory for threads used to run tests.
```

src/share/classes/com/sun/javatest/regtest/agent/TestThreadFactory.java
```diff
@@ -0,0 +1,84 @@
...
+ * The implementation of this interface might provide user-defined threads for test execution.
```
jonathan-gibbons, line 40: Change `The` to `An`

src/share/classes/com/sun/javatest/regtest/agent/TestThreadFactory.java
jonathan-gibbons, line 39: Line 39, `jtreg creates a new thread...`

If I understand you correctly, this might be better as:

```
By default, jtreg creates a new thread for each test using {@code ...}, but this may be overridden by providing
an implementation of this interface, which might provide user-defined threads for test execution.
```

src/share/classes/com/sun/javatest/regtest/tool/i18n.properties
```diff
@@ -171,6 +171,12 @@ help.main.ignore.error.desc=(Default.) Execute the actions up to the @ignore tag
     then give an "Error" result.
 help.main.ignore.run.desc=Run the test, as though the @ignore tag were not present.
 help.main.l.desc=List the tests that would be executed instead of executing them.
+help.main.ttf.arg=<classname>
+help.main.ttf.desc=Specifies the class to create thread factory for test execution. \
+    The class should implement a j.u.c.TestFactory interface.
```
jonathan-gibbons, line 176: change "a" to "the" and spell out `j.u.c` in full

src/share/classes/com/sun/javatest/regtest/agent/TestThreadFactoryHelper.java
```diff
@@ -0,0 +1,80 @@
...
+            SearchPath classpath = new SearchPath(path);
+            List<URL> urls = new ArrayList<>();
+            for (Path f : classpath.asList()) {
+                try {
+                    urls.add(f.toUri().toURL());
+                } catch (MalformedURLException e) {
+                }
+            }
+            loader = new URLClassLoader(urls.toArray(new URL[urls.size()]), loader);
+        }
```
jonathan-gibbons, line 68: In the `othervm` case, this code will presumably just be called once, but in the `agentvm` case, won't it be called once per test, and since the class loader is never closed, (see [URLCLassLoader.close](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/net/URLClassLoader.html#close())) won't that lead to a leak of any resources that might be involved ... especially any zip/jar files that might be opened?

lmesnik: I have updated the code to close the classloader. The jtreg doesn't have now any embedded implementations so path never could be null, I suppose.

jonathan-gibbons (approved):

PR #172, opened 2023-10-10
7903526: jtreg should handle all exceptions

lmesnik:
Currently, jtreg executes a test in MainThreadGroup and checks uncaught exceptions for this group. So it handles all exceptions if the test doesn't create new thread groups.
However, the virtual threads don't belong to MainThreadGroup and jtreg silently ignores all exceptions thrown by virtual threads.

The testing shows that we have already some problems in jdk/jdk and probably in UR releases. There are 13 test failures in tier1 and several tens in the execution of tests that use secutiry-manager and run with a virtual thread.

It wonder if it makes sense to have this check conditional using some java property and start fixing tests after jtreg is release and smoothly switch execution to the new version.

---------
### Progress

### Issue
 * [CODETOOLS-7903526](https://bugs.openjdk.org/browse/CODETOOLS-7903526): jtreg should handle all exceptions (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/172/head:pull/172` \
`$ git checkout pull/172`

`$ git checkout pull/172` \
`$ git pull https://git.openjdk.org/jtreg.git pull/172/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 172`

`$ git pr show -t 172`

</details>

<a href="https://git.openjdk.org/jtreg/pull/172.diff">https://git.openjdk.org/jtreg/pull/172.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/172#issuecomment-1756242348)

### DIFF
--- a/src/share/classes/com/sun/javatest/regtest/agent/MainActionHelper.java
+++ b/src/share/classes/com/sun/javatest/regtest/agent/MainActionHelper.java
@@ -61,7 +61,7 @@ public class MainActionHelper extends ActionHelper {
     private String testThreadFactoryPath;
     private OutputHandler outputHandler;
 
-    private Throwable globalUncaughtThrowable;
+    private volatile Throwable globalUncaughtThrowable;
 
     MainActionHelper(String testName) {
         this.testName = testName;
--- a/src/share/classes/com/sun/javatest/regtest/agent/MainActionHelper.java
+++ b/src/share/classes/com/sun/javatest/regtest/agent/MainActionHelper.java
@@ -61,6 +61,8 @@ public class MainActionHelper extends ActionHelper {
     private String testThreadFactoryPath;
     private OutputHandler outputHandler;
 
+    private Throwable globalUncaughtThrowable;
+
     MainActionHelper(String testName) {
         this.testName = testName;
     }
@@ -200,6 +202,15 @@ public class MainActionHelper extends ActionHelper {
             // Main and Thread are same here
             // RUN JAVA IN ANOTHER THREADGROUP
             AgentVMThreadGroup tg = new AgentVMThreadGroup(err, MSG_PREFIX, timeoutFactor);
+            Thread.setDefaultUncaughtExceptionHandler((t, e) -> {
+                synchronized (this) {
+                    if (e instanceof ThreadDeath)
+                        return;
+                    if (globalUncaughtThrowable == null) {
+                        globalUncaughtThrowable = e;
+                    }
+                }
+            });
             Thread t;
             if (testThreadFactory == null) {
                 t = new Thread(tg, avmr);
@@ -247,10 +258,11 @@ public class MainActionHelper extends ActionHelper {
                 }
             }
 
-            if (((avmr.t != null) || (tg.uncaughtThrowable != null)) && (error == null)) {
+            if (((avmr.t != null) || (tg.uncaughtThrowable != null) || (globalUncaughtThrowable != null))
+                    && (error == null)) {
                 if (avmr.t == null) {
-                    error = tg.uncaughtThrowable;
-                } else {
+                    error = tg.uncaughtThrowable != null ? tg.uncaughtThrowable : globalUncaughtThrowable;
+                }  else {
                     error = avmr.t;
                 }
                 if (SKIP_EXCEPTION.equals(error.getClass().getName())) {
@@ -286,6 +298,7 @@ public class MainActionHelper extends ActionHelper {
         } finally {
             out.close();
             err.close();
+            Thread.setDefaultUncaughtExceptionHandler(null);
             status = saved.restore(testName, status);
         }
 
--- a/src/share/classes/com/sun/javatest/regtest/agent/MainWrapper.java
+++ b/src/share/classes/com/sun/javatest/regtest/agent/MainWrapper.java
@@ -39,6 +39,8 @@ public class MainWrapper
     public static String TEST_THREAD_FACTORY = "jtreg.test.thread.factory";
     public static String TEST_THREAD_FACTORY_PATH = "jtreg.test.thread.factory.path";
 
+    private static Throwable globalUncaughtThrowable = null;
+
     public static void main(String[] args) {
         String moduleName;
         String className;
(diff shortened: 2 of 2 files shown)


src/share/classes/com/sun/javatest/regtest/agent/MainActionHelper.java
```diff
@@ -247,10 +258,11 @@ public AStatus runClass() {
                 }
             }
 
-            if (((avmr.t != null) || (tg.uncaughtThrowable != null)) && (error == null)) {
+            if (((avmr.t != null) || (tg.uncaughtThrowable != null) || (globalUncaughtThrowable != null))
+                    && (error == null)) {
                 if (avmr.t == null) {
-                    error = tg.uncaughtThrowable;
-                } else {
+                    error = tg.uncaughtThrowable != null ? tg.uncaughtThrowable : globalUncaughtThrowable;
```
AlanBateman, line 264: Are you missing synchronisation here? The VM-side UHE is set while holding the lock for the MainActionHelper instance whereas help you read without any synchronisation. If you make globalUncaughtThrowable volatile then it would be okay, meaning keep the synchronisation in the setter code, read from elsewhere without synchronisation.

src/share/classes/com/sun/javatest/regtest/agent/MainWrapper.java
```diff
@@ -70,6 +72,18 @@ public static void main(String[] args) {
...
+
+        Thread.setDefaultUncaughtExceptionHandler((t, e) -> {
+            synchronized (MainWrapper.class) {
+                if (e instanceof ThreadDeath)
+                    return;
+                e.printStackTrace(System.err);
+                if (globalUncaughtThrowable == null) {
+                    globalUncaughtThrowable = e;
+                }
+                AStatus.failed(MAIN_THREW_EXCEPT + e).exit();
```
jaikiran, line 84: I think we shouldn't be failing and exiting here because the uncaught exception might have been thrown when the test is still in progress. Just keeping track of the uncaught exception(s) I think should be enough, so that that state can then be used after the test completes, to decide whether to fail the test. This is then similar to what we do in the case where no test thread factory is set.

src/share/classes/com/sun/javatest/regtest/agent/MainWrapper.java
```diff
@@ -70,6 +72,18 @@ public static void main(String[] args) {
 
         // RUN JAVA IN ANOTHER THREADGROUP
         MainThreadGroup tg = new MainThreadGroup();
+
+        Thread.setDefaultUncaughtExceptionHandler((t, e) -> {
```
jaikiran, line 76: I think if we do decide to set a JVM level default uncaught exception handler then it has to be done only for the case where the `Thread` instance returned by the test thread factory (if there is one) is a virtual thread. So something like:

```java
if (t.isVirtual()) {
  Thread.setDefaultUncaughtExceptionHandler(.....)
}
```
That way we don't interfere with uncaught exception handlers when platform thread is being used.

jaikiran: Finally, the major issue with setting a JVM level uncaught exception handler is that, tests will no longer be able to expect the `Thread.getDefaultUncaughtExceptionHandler()` to ever be `null`, even if they launch in `/othervm` mode. I think that's a problem. I can't recollect if jtreg testing framework has other similar cases where it interferes with the tests even with the usage of `/othervm`.

src/share/classes/com/sun/javatest/regtest/agent/MainActionHelper.java
lmesnik: Thanks, added volatile.

src/share/classes/com/sun/javatest/regtest/agent/MainWrapper.java
lmesnik: No, the plan is to catch all unhandled exceptions. The virtual threads are only a particular case but might be the most common. However, I think we shouldn't silently ignore any of the unhandled exceptions. Some exceptions are already handled using ThreadGroup, and the fix doesn't break this, just add more handling.

lmesnik: > Finally, the major issue with setting a JVM level uncaught exception handler is that, tests will no longer be able to expect the Thread.getDefaultUncaughtExceptionHandler() to ever be null, even if they launch in /othervm mode. I think that's a problem. I can't recollect if jtreg testing framework has other similar cases where it interferes with the tests even with the usage of /othervm.

I don't think it is a major issue.  It might be useful to document this, so tests don't expect global UHE to be null.  They could re-write UHE if they need and don't need to preserve/restore the original one.

src/share/classes/com/sun/javatest/regtest/agent/MainWrapper.java
lmesnik: I copied the handler from MainThreadGroup::uncaughtException. I want to keep them consistent. (Eventually, we are going to get rid of MainGroup.) Probably, it is needed to file a bug not to fail in the MainThreadGroup::uncaughtException and check exceptions after the test passes.

jaikiran: Hello Leonid,

> I copied the handler from MainThreadGroup::uncaughtException. I want to keep them consistent.

I hadn't noticed MainThreadGroup does it this way. Having it consistent then sounds good to me.

src/share/classes/com/sun/javatest/regtest/agent/MainWrapper.java
jaikiran: > However, I think we shouldn't silently ignore any of the unhandled exceptions. Some exceptions are already handled using ThreadGroup

Do you mean there are uncaught exceptions that aren't currently being handled/reported by jtreg when using platform thread?

lmesnik: I hope that we don't have. Most of test platform threads should inherit MainGroup. However, I don't know if it is true for all pools and executors implementations. And there are JDK internal threads (Managment, JVMCi, Debugger, UI, JFR and other). I hope that they handle exceptions, but haven't checked really all of them. The goal of this fix to ensure that we don't miss any of them if they were thrown during test execution.

lmesnik:
any ideas about moving forward with this PR?

PR #173, opened 2023-11-13
7903580: Allow for re-attempting agent creation when an attempt fails

jaikiran:
Can I please get a review of this change which proposes to implement the enhancement request noted in https://bugs.openjdk.org/browse/CODETOOLS-7903580?

Several times in our CI instances we have noticed that a test (action) fails because it is unable to create an `Agent` instance due to the socket connection not being established between the `AgentServer` and the `ServerSocket` running within the jtreg process. This causes the test itself to fail with `Error. Cannot get VM for test: java.net.SocketTimeoutException: Accept timed out`. It has also been noticed that most of the times this is intermittent and subsequent attempt for a different test (action) passes. 

The proposed change here introduces a new configuration parameter `--agent-attempts` under "Agent Pool Options", which allows for configuring the maximum number of attempts that are allowed for getting an agent for an action. This is an optional parameter and by default it has been set to a value of `2`, which by default then allows the agent creation to be retried once if the previous attempt fails. This then means that existing installations/usages of jtreg need not set a value for `--agent-attempts` to enroll for this feature. A value of `1` for this parameter implies that the agent creation attempt will be done only once, which is what currently happens in the absence of this proposed feature.

Inability to obtain an agent, for whatever reason, after attempting `--agent-attempts` times continues to result in test (action) failure, like it does today.

Some additional minor logging related changes have been done too, to help debugging some of the current observed failures.

This change has been tested in the following manners:

1. This modified build of jtreg has been used to run tier1, tier2 and tier3 of JDK mainline and it has been verified that nothing regresses with this change. That however doesn't mean that agent creation re-attempt logic was tested in these runs (since the first attempt never failed).
2. Locally this build was jtreg was additionally modified to selectively not connect to the `ServerSocket`'s port thus simulating a "Accept timed out" exception. That change then verified that the re-attempt logic does indeed correctly kick in and the test action passes after picking up the newly created agent.
3. `--agent-attempts 1` was passed to the above modified build (from step 2) to verify that the re-attempt isn't attempted when `--agent-attempts 1` and the test fails as currently with "test result: Error. Cannot get VM for test: java.net.SocketTimeoutException: Accept timed out"

---------
### Progress

### Issue
 * [CODETOOLS-7903580](https://bugs.openjdk.org/browse/CODETOOLS-7903580): Allow for re-attempting agent creation when an attempt fails (**Enhancement** - P4)

### Reviewers
 * [Christian Stein](https://openjdk.org/census#cstein) (@sormuras - Committer) ⚠️ Review applies to [3d08cdfe](https://git.openjdk.org/jtreg/pull/173/files/3d08cdfe4bcce041e37cca0222cf00feb4512067)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/173/head:pull/173` \
`$ git checkout pull/173`

`$ git checkout pull/173` \
`$ git pull https://git.openjdk.org/jtreg.git pull/173/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 173`

`$ git pr show -t 173`

</details>

<a href="https://git.openjdk.org/jtreg/pull/173.diff">https://git.openjdk.org/jtreg/pull/173.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/173#issuecomment-1808217547)

### DIFF
--- a/src/share/classes/com/sun/javatest/regtest/exec/Agent.java
+++ b/src/share/classes/com/sun/javatest/regtest/exec/Agent.java
@@ -808,6 +808,7 @@ public class Agent {
                        String testThreadFactoryPath)
                 throws Fault {
             final int numAttempts = this.numAgentSelectionAttempts;
+            assert numAttempts > 0 : "unexpected agent selection attempts: " + numAttempts;
             Agent.Fault toThrow = null;
             for (int i = 1; i <= numAttempts; i++) {
                 try {
@@ -1022,8 +1023,7 @@ public class Agent {
         private float timeoutFactor = 1.0f;
         private int maxPoolSize;
         private Duration idleTimeout;
-        // default is 1 i.e. we don't re-attempt a failed agent selection
-        private int numAgentSelectionAttempts = 1;
+        private int numAgentSelectionAttempts;
     }
 
     static class Stats {
--- a/src/share/classes/com/sun/javatest/regtest/exec/Agent.java
+++ b/src/share/classes/com/sun/javatest/regtest/exec/Agent.java
@@ -37,7 +37,6 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.PrintStream;
 import java.io.PrintWriter;
-import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.net.InetSocketAddress;
 import java.net.ServerSocket;
@@ -98,7 +97,7 @@ public class Agent {
         Method pidMethod = null;
         try {
             pidMethod = Process.class.getDeclaredMethod("pid"); // only available in Java 9+
-        } catch (NoSuchMethodException | SecurityException e) {
+        } catch (Exception e) {
             pidMethod = null;
         }
         PID_METHOD = pidMethod;
@@ -605,7 +604,7 @@ public class Agent {
         }
         try {
             return (long) PID_METHOD.invoke(process);
-        } catch (IllegalAccessException | InvocationTargetException e) {
+        } catch (Exception e) {
             return UNKNOWN_PID;
         }
     }
@@ -724,10 +723,6 @@ public class Agent {
             logger = Logger.instance(params);
         }
 
-        void log(final String message) {
-            this.logger.log(null, "POOL: " + message);
-        }
-
         /**
          * Sets the policy file to be used for agents created for this pool.
          *
@@ -769,7 +764,7 @@ public class Agent {
         /**
          * Sets the maximum attempts to create or obtain an agent VM
          * @param numAttempts number of attempts
-         * @throws IllegalArgumentException if {@code numAttempts} is lesser than {@code 1}
+         * @throws IllegalArgumentException if {@code numAttempts} is less than {@code 1}
          */
         public void setNumAgentSelectionAttempts(final int numAttempts) {
             if (numAttempts < 1) {
@@ -778,10 +773,6 @@ public class Agent {
             }
             this.numAgentSelectionAttempts = numAttempts;
             logger.log(null, "POOL: agent selection attempts: " + numAttempts);
-        }
-
-        int getNumAgentSelectionAttempts() {
-            return this.numAgentSelectionAttempts;
         }
 
         /**
@@ -799,7 +790,47 @@ public class Agent {
          * @return the agent
          * @throws Fault if there is a problem obtaining a suitable agent
          */
-        synchronized Agent getAgent(File dir,
+        Agent getAgent(File dir,
+                       JDK jdk,
+                       List<String> vmOpts,
+                       Map<String, String> envVars,
+                       String testThreadFactory,
+                       String testThreadFactoryPath)
+                throws Fault {
+            final int numAttempts = this.numAgentSelectionAttempts;
+            Agent.Fault toThrow = null;
+            for (int i = 1; i <= numAttempts; i++) {
+                try {
+                    if (i != 1) {
+                        logger.log(null, "POOL: re-attempting agent creation, attempt number " + i);
+                    }
+                    return doGetAgent(dir, jdk, vmOpts, envVars, testThreadFactory,
+                            testThreadFactoryPath);
+                } catch (Agent.Fault f) {
+                    logger.log(null, "POOL: agent creation failed due to " + f.getCause());
+                    // keep track of the fault and reattempt to get an agent if within limit
+                    if (toThrow == null) {
+                        toThrow = f;
+                    } else {
+                        // add the previous exception as a suppressed exception
+                        // of the current one
+                        if (toThrow.getCause() != null) {
+                            f.addSuppressed(toThrow.getCause());
+                        }
+                        toThrow = f;
+                    }
+                    if (i == numAttempts || !(f.getCause() instanceof IOException)) {
+                        // we either made enough attempts or we failed due to a non IOException.
+                        // In either case we don't attempt to create an agent again and instead
+                        // throw the captured failure(s)
+                        throw toThrow;
+                    }
+                }
+            }
+            throw new AssertionError("should not reach here");
+        }
+
+        synchronized Agent doGetAgent(File dir,
                                     JDK jdk,
                                     List<String> vmOpts,
                                     Map<String, String> envVars,
@@ -822,13 +853,8 @@ public class Agent {
                 stats.reuse(a);
             } else {
                 logger.log(null, "POOL: Creating new agent");
-                try {
-                    a = new Agent(dir, jdk, vmOpts, envVars, policyFile, timeoutFactor, logger,
-                            testThreadFactory, testThreadFactoryPath);
-                } catch (Fault f) {
-                    logger.log(null, "POOL: Agent creation failed due to " + f.getCause());
-                    throw f;
-                }
+                a = new Agent(dir, jdk, vmOpts, envVars, policyFile, timeoutFactor, logger,
+                        testThreadFactory, testThreadFactoryPath);
                 stats.add(a);
             }
 
@@ -986,8 +1012,8 @@ public class Agent {
         private float timeoutFactor = 1.0f;
         private int maxPoolSize;
         private Duration idleTimeout;
-        // default is 2 i.e. we retry a failed agent selection once
-        private int numAgentSelectionAttempts = 2;
+        // default is 1 i.e. we don't re-attempt a failed agent selection
+        private int numAgentSelectionAttempts = 1;
     }
 
     static class Stats {
--- a/src/share/classes/com/sun/javatest/regtest/exec/Agent.java
+++ b/src/share/classes/com/sun/javatest/regtest/exec/Agent.java
@@ -37,6 +37,8 @@ import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.PrintStream;
 import java.io.PrintWriter;
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
 import java.net.InetSocketAddress;
 import java.net.ServerSocket;
 import java.net.Socket;
@@ -90,6 +92,18 @@ public class Agent {
     static final boolean showAgent = Flags.get("showAgent");
     static final boolean traceAgent = Flags.get("traceAgent");
 
+    private static final long UNKNOWN_PID = -1;
+    private static final Method PID_METHOD;
+    static {
+        Method pidMethod = null;
+        try {
+            pidMethod = Process.class.getDeclaredMethod("pid"); // only available in Java 9+
+        } catch (NoSuchMethodException | SecurityException e) {
+            pidMethod = null;
+        }
+        PID_METHOD = pidMethod;
+    }
+
     /**
      * Start a JDK with given JVM options.
      */
@@ -127,8 +141,9 @@ public class Agent {
             // is platform-specific, and Solaris has it on by default.
             ss.setReuseAddress(false);
             ss.bind(new InetSocketAddress(/*port:*/ 0), /*backlog:*/ 1);
+            final int port = ss.getLocalPort();
             cmd.add(AgentServer.PORT);
-            cmd.add(String.valueOf(ss.getLocalPort()));
+            cmd.add(String.valueOf(port));
 
             if (timeoutFactor != 1.0f) {
                 cmd.add(AgentServer.TIMEOUTFACTOR);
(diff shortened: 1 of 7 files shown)


jonathan-gibbons:
> The proposed change here introduces a new configuration parameter --agent-attempts under "Agent Pool Options", which allows for configuring the maximum number of attempts that are allowed for getting an agent for an action. This is an optional parameter and by default it has been set to a value of 2, which by default then allows the agent creation to be retried once if the previous attempt fails. 

A different strategy might be to set the default to 1, to retain existing behavior by default, but use the option to raise the value when run with the makefiles, such as on Mach5.

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
```diff
@@ -90,6 +92,18 @@ public static class Fault extends Exception {
     static final boolean showAgent = Flags.get("showAgent");
     static final boolean traceAgent = Flags.get("traceAgent");
 
+    private static final long UNKNOWN_PID = -1;
+    private static final Method PID_METHOD;
+    static {
+        Method pidMethod = null;
+        try {
+            pidMethod = Process.class.getDeclaredMethod("pid"); // only available in Java 9+
+        } catch (NoSuchMethodException | SecurityException e) {
```
jonathan-gibbons, line 101: It can throw `UnsupportedOperationException`
https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/Process.html#pid()

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
```diff
@@ -730,6 +766,24 @@ public void setMaxPoolSize(int size) {
             logger.log(null, "POOL: max pool size: " + maxPoolSize);
         }
 
+        /**
+         * Sets the maximum attempts to create or obtain an agent VM
+         * @param numAttempts number of attempts
+         * @throws IllegalArgumentException if {@code numAttempts} is lesser than {@code 1}
```
jonathan-gibbons, line 772: s/lesser/less/

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
```diff
@@ -927,6 +986,8 @@ private static String getKey(File dir, JDK jdk, List<String> vmOpts) {
         private float timeoutFactor = 1.0f;
         private int maxPoolSize;
         private Duration idleTimeout;
+        // default is 2 i.e. we retry a failed agent selection once
+        private int numAgentSelectionAttempts = 2;
```
jonathan-gibbons, line 990: The default should be 1, meaning that the new behavior is opt-in, not opt-out.

You can always set a different default in the layers of software wrapping `jtreg`.

src/share/classes/com/sun/javatest/regtest/exec/RegressionScript.java
```diff
@@ -1229,10 +1230,39 @@ Agent getAgent(JDK jdk, SearchPath classpath, List<String> testVMOpts,
...
-        return agent;
+        final int numAttempts = p.getNumAgentSelectionAttempts();
+        assert numAttempts >= 1 : "invalid agent selection attempts: " + numAttempts;
+        Agent.Fault toThrow = null;
+        for (int i = 1; i <= numAttempts; i++) {
+            try {
+                if (i != 1) {
+                    p.log("Re-attempting agent creation, attempt number " + i);
+                }
+                Agent agent = p.getAgent(absTestScratchDir().toFile(), jdk, vmOpts.toList(),
```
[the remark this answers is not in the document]
jonathan-gibbons: Can you not make this blob of code be a static or unsynchronized method in `Agent.Pool`?

Approximately speaking, if you can do this code here, you should be able to do it there.

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
```diff
@@ -459,6 +459,22 @@ public void process(String opt, String arg) throws BadArgs {
             }
         },
 
+        new Option(GNU, AGENT_POOL, null, "--agent-attempts") {
```
jonathan-gibbons, line 462: It might be good to add a FAQ entry for this option as well.

Q. `jtreg` has trouble starting agents; what can I do?
A. This can happen because _<fill-in-the-blank>_. Use the `--sgents-attempts` option to retry creating a agent up to a given number of times.

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
jonathan-gibbons, line 990: The limit of 2 is set here but also later on.

src/share/classes/com/sun/javatest/regtest/tool/i18n.properties
```diff
@@ -294,6 +294,15 @@ help.pool.pool_idle_timeout.arg=<number>
...
      The time, in seconds, before an idle VM in the pool is \
      automatically closed.
+help.pool.agent_attempts.arg=<number>
+help.pool.agent_attempts.desc=\
+    The maximum number of times the pool will be queried for a agent VM \
+    when running an action. A value of 1 will mean that no retry will be \
+    attempted if either a new agent VM creation fails or an existing one \
+    couldn't be obtained from the pool. Similarly, a value of 2 will imply \
+    that if the agent VM creation fails or an existing one couldn't be \
+    obtained from the pool, then 1 retry will be attempted. \
```
jonathan-gibbons, line 304: The text about "queries" seems "weird".  Suggest:

The number of attempts jtreg will make to access an agent with the desired characteristics, either by creating a new one, or obtaining one from the pool of reusable VMs. The minimum value, and the default value, is 1.

jonathan-gibbons (changes requested):

src/share/classes/com/sun/javatest/regtest/exec/RegressionScript.java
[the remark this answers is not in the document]
jaikiran: Hello Jon,
> Can you not make this blob of code be a static or unsynchronized method in Agent.Pool?

That's a good idea. It makes it much more cleaner and internal to the pool implementation itself. I've now moved this newly introduced code into a package private `Pool.getAgent(...)` method without any synchronization. This method does the necessary re-attempts of getting an agent. The previous existing `synchronized` `getAgent(...)` has just been renamed to `private synchronized doGetAgent(...)` - no other changes have been introduced in that method.

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
jaikiran: Good catch. Given that the reason I introduced the pid() call was for better debuggability, I have now updated to PR to catch the broad `Exception` type, both at the place where we reflectively check for the presence of `pid` method as well as where we reflectively call the `pid()` method, to ignore any exceptions and return `-1` in such cases.

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
jaikiran: Fixed.

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
jaikiran: I've updated the PR to use the default as 1.

jaikiran: I did not understand this review comment. With the updated changes in this PR, the default value now is set to 1, here too. Did you mean something more?

src/share/classes/com/sun/javatest/regtest/tool/i18n.properties
jaikiran: Thank you for providing this text. I agree it's much more concise and provides all the necessary details. I've updated the PR to use this text.

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
jaikiran: Hello Jon, I've now updated the PR to include a FAQ entry too.

jaikiran:
I've been using a jtreg version containing this patch for around 10 days now in our CI setup for my personal JDK jobs. So far I haven't seen any regression caused by this change.

sormuras (approved):
Look

msheppar:
For my own edification, I'm curious as to why the Process::pid method needs to be called reflectively when pb.start() has returned a a reference to the newly created process?
168             process = pb.start();
169             final long pid = getPid(process);

sormuras:
> `pidMethod = Process.class.getDeclaredMethod("pid"); // only available in Java 9+`

Agents need to be able to run on older Java version as documented in `package-info.java` of `package com.sun.javatest.regtest.agent;`: 

https://github.com/openjdk/jtreg/blob/a12c75c56a2aa3f39c3f408d2468fa953132bb6e/src/share/classes/com/sun/javatest/regtest/agent/package-info.java#L31-L40

Caveat: those numbers might be out of date and are in need of a refresh.

msheppar:
I have not studied the solution in depth, so bear with me on this questions, in the getAgent call flow, I can see two main sources of an IOException, the ProcessBuilder::start and the ServerSocket::accept. If the accept is the source of the IOException and the initiation of a retry, what happens to the instantiated process, which could still be potentially alive. Do you check that it is alive and (forcibly) terminate it if necessary ?

jaikiran:
Hello Mark,

> For my own edification, I'm curious as to why the Process::pid method needs to be called reflectively when pb.start()

I see Christian already answered this part - for jtreg to be usable in Java 8, we cannot statically use APIs that have been introduced in later versions. `Process::pid()` was introduced in Java 9.

> If the accept is the source of the IOException and the initiation of a retry, what happens to the instantiated process, which could still be potentially alive. Do you check that it is alive and (forcibly) terminate it if necessary ?

When an `Agent` is constructed, the `ProcessBuilder` launches the `AgentServer` class which uses `new Socket(....)` to connect to the `ServerSocket` which is hosted by the jtreg process. If the `accept()` times out or fails due to an IOException, then the `ServerSocket` is closed. Once this happens, the jtreg process itself doesn't do any checks on the `AgentServer` process since the `Agent` instance itself couldn't be constructed. However, since the `AgentServer` process is in the `new Socket(...)` code waiting for the connection to be accepted, that call would then fail (there's no server port listening for connections, keeping aside cases like some other process now listening on this port) and that would then trigger the AgentServer process to exit.

Having said that, there's a case to be made that if the construction of the `Agent` class fails, then the created resources (the AgentServer `process` in this case) should also be destroyed. I'll take a look to see if doing that will cause any issues.

msheppar:
cheers (C & J) for the clarifications

jonathan-gibbons (changes requested):
Mostly good; nearly there, but some questions/changes remain

src/share/doc/javatest/regtest/faq.md
```diff
@@ -699,6 +699,24 @@ if the default mode was not used.  For example,
...
+agent VM and instead the failure is propagated as a test action failure and
+that action gets reported as a failed.
+
+This default behaviour can be overridden by passing the `--agent-attempts` option
+to `jtreg` command. This option takes an integer value which represents the number
+of attempts jtreg is allowed to make when attempting to get a agent for a test
+action. By default, the value of this option is `1`, implying jtreg will not
+re-attempt a failed attempt. Passing a higher value for this option will allow
+jtreg to re-attempt a failed attempt. For example, a value of `2` will allow jtreg
+to re-attempt once for each failed attempt.
```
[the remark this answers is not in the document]
jaikiran: Done. I've updated this section of the FAQ, to use JTReg, except for one place where the sentence refers to the `jtreg` command.

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
```diff
@@ -2330,9 +2330,9 @@ public enum ReportMode { NONE, EXECUTED, FILES, ALL_EXECUTED, ALL }
     private String testThreadFactoryPathArg;
     private int maxPoolSize = -1;
     private Duration poolIdleTimeout = Duration.ofSeconds(30);
-    // number of attempts to get an agent for an action. we default to 2, to allow
-    // for retrying a failed attempt once
-    private int numAgentSelectionAttempt = 2;
+    // number of attempts to get an agent for an action. we default to 1, which implies
+    // by default we don't re-attempt on a failure
+    private int numAgentSelectionAttempt = 1;
```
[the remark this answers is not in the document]
jaikiran: Hello Jon, I've updated the PR to set this default at a single place and also used a static final constant for the default value.

jaikiran:
Please keep open, expecting further reviews.

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
```diff
@@ -90,12 +91,25 @@ public static class Fault extends Exception {
     static final boolean showAgent = Flags.get("showAgent");
     static final boolean traceAgent = Flags.get("traceAgent");
```
jonathan-gibbons, line 93: (minor) suggest adding a comment that this is to allow running `jtreg` on older JDKs that do not support the `pid` method.

src/share/doc/javatest/regtest/faq.md
```diff
@@ -699,6 +699,24 @@ if the default mode was not used.  For example,
...
+Agent VM creation involves launching a process and communicating with it over
+a socket. The initial handshake between the newly launched process and JTReg
+can sometimes timeout if the system is under heavy load. This then causes the
+agent creation to fail. By default, JTReg does not re-attempt creation of the
+agent VM and instead the failure is propagated as a test action failure and
+that action gets reported as a failed.
+
+This default behaviour can be overridden by passing the `--agent-attempts` option
+to `jtreg` command. This option takes an integer value which represents the number
+of attempts JTReg is allowed to make when attempting to get a agent for a test
```
jonathan-gibbons, line 714: (minor) remove `JTReg is allowed` so that it reads "the number of attempts to make..."

jonathan-gibbons (approved):
Nice.

The FAQ entry is a nice addition, to explain the typical reason that launching an agent might fail, and so why it is reasonable to have a retry count.

src/share/doc/javatest/regtest/faq.md
jaikiran: Done, updated the PR to remove that part.

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
jaikiran: Done - I've updated the PR to include a code comment for this line.

jaikiran:
Thank you both Jon and Christian for these reviews. I'll go ahead and integrate this today or tomorrow (once I run one more run in CI).

jonathan-gibbons (approved):

jaikiran:
No issues noticed with the recent run of CI job with these changes. I'll go ahead with the integration now (requires a sponsor).

PR #181, opened 2024-02-09
7903659 Verify Problemlist contents

LudwikJaniuk:
This change introduces the flag `--verify-exclude` (short form `-i`) which will run additional checks on the problemlists passes through `-exclude`. If any of the checks fails, jtreg prints a helpful message and refuses to start running tests.

---------
### Progress

### Issue
 * [CODETOOLS-7903659](https://bugs.openjdk.org/browse/CODETOOLS-7903659): Verify Problemlist contents (**Enhancement** - P3)

### Reviewers
 * [Jonathan Gibbons](https://openjdk.org/census#jjg) (@jonathan-gibbons - **Reviewer**) ⚠️ Review applies to [9a9a5f21](https://git.openjdk.org/jtreg/pull/181/files/9a9a5f21a07385495f48d11b83efe08e80dd50fb)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/181/head:pull/181` \
`$ git checkout pull/181`

`$ git checkout pull/181` \
`$ git pull https://git.openjdk.org/jtreg.git pull/181/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 181`

`$ git pr show -t 181`

</details>

<a href="https://git.openjdk.org/jtreg/pull/181.diff">https://git.openjdk.org/jtreg/pull/181.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/181#issuecomment-1952603044)


LudwikJaniuk:
I feel confident enough to take this out of draft status, but am eagerly awaiting feedback. The style of writing tests in this repo is unfamiliar to me, so I might not have done it in the most elegant way. I hope to have basically covered the functionality I'm adding though.

sormuras:
It's good to have such a verification in `jtreg`/`jtharness` soon.

I wonder whether an existing option should be (re-) used to trigger the verification process. `jtreg.jar --help verify` prints:
```
General Options

    -c | -check     Verify correctness of test descriptions. Does NOT run tests.
    -xml | --xml:verify
                    Create ant/junit xml files into the workDir. Optionally
                    verify if the file is well-formed.
```

Perhaps `-c | -check` can be amended to also verify the correctness of _exlude files_ and maybe later _match files_, too?
```
Test Selection Options
                These options can be used to refine the set of tests to be
                executed.

    -match:<file>   Provide a file specifying tests that can be run (inverse of
                    -exclude)
```

src/share/classes/com/sun/javatest/regtest/tool/ExcludeFileVerifier.java
```diff
@@ -0,0 +1,118 @@
...
+        try {
+            BufferedReader br = new BufferedReader(new FileReader(file));
+            String line = null;
+            int n = 0;
+            while ((line = br.readLine()) != null)
+            {
+                n++;
+                if (lineIsComment(line.trim())) continue;
+                for(Check c : checks) {
+                    if(!c.check(line.trim())) {
```
jonathan-gibbons, line 40: Code style issues.   
* Space between keyword and left-paren
* `{` at end of line, not beginning of next

src/share/classes/com/sun/javatest/regtest/tool/ExcludeFileVerifier.java
```diff
@@ -0,0 +1,118 @@
...
+                        System.out.println(file.getAbsolutePath() + " line " + n + " is invalid. Reason:");
+                        System.out.println(c.description());
+                        System.out.println("Line contents:");
+                        System.out.println("--------------");
+                        System.out.println(line);
+                        System.out.println("--------------");
```
jonathan-gibbons, line 46: Don't use `System.out` directly like this; pass in a `PrintWriter` to be used. 

The main `jtreg``Tool` class sets up two stream to be used, `out` and `err`.  
Generally, 
* `standard output` / `System.out` / `sysOut` / `stdOut` should be used for the expected output of specific options, like command-line help.
* `standard error` / `System.err` / `sysErr` / `stdErr` should be used for diagnostic output when issues are found performing any requested operation.

src/share/classes/com/sun/javatest/regtest/tool/ExcludeFileVerifier.java
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    abstract class Check {
+        public abstract String description();
+        public abstract boolean check(String line);
+    }
+
+    class LineFormatCheck extends Check {
+        private static final String commalist = "([\\w-]+)(,[\\w-]+)*";
+        private static final String p = "\\S+\\s+" + commalist + "\\s" + commalist + ".*";
```
jonathan-gibbons, line 77: Hmm, I see what you're doing, but generally it is better to use `Pattern.compile` for regular expressions.

src/share/classes/com/sun/javatest/regtest/tool/ExcludeFileVerifier.java
```diff
@@ -0,0 +1,118 @@
...
+        public String description() {
+            return "Must follow: <test-name> <bugid>(,<bugid>)* <platform>(,<platform>)* <description>";
+        }
+
+        public boolean check(String line) {
+            return Pattern.matches(p, line);
```
jonathan-gibbons, line 83: Not a huge deal, but this will compile the regular expression every time it is called. It is better to compile the regex once, using `Pattern.compile` and then create a `Matcher` from the `Pattern`.

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
```diff
@@ -642,6 +643,13 @@ public void process(String opt, String arg) {
             }
         },
 
+        new Option(NONE, MAIN, null, "-ve", "-verifyexclude") {
```
jonathan-gibbons, line 646: I might have gone for `--verify-exclude` but this is OK

jonathan-gibbons (changes requested):
Very good for a first attempt.
Some minor suggestions given.

jonathan-gibbons:
Please fix the Description of the PR -- the name of the option there does not match that in the code. The one in the code is better.

> This change introduces the flag --excludeverify

jonathan-gibbons:
> The style of writing tests in this repo is unfamiliar to me

Yes, sorry about that; you should have seen the mental contortions when we tried to use JavaTest to test JavaTest all those many many years ago!

test/excludeverify/ExcludeVerifyTest.gmk
```diff
@@ -0,0 +1,149 @@
+#
+# Copyright (c) 2012, 2022, Oracle and/or its affiliates. All rights reserved.
+# DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+#
+# This code is free software; you can redistribute it and/or modify it
+# under the terms of the GNU General Public License version 2 only, as
+# published by the Free Software Foundation.  Oracle designates this
+# particular file as subject to the "Classpath" exception as provided
+# by Oracle in the LICENSE file that accompanied this code.
```
jonathan-gibbons, line 9: Please remove `Oracle designates ... this code`.

test/excludeverify/ExcludeVerifyTest.gmk
```diff
@@ -0,0 +1,149 @@
+#
+# Copyright (c) 2012, 2022, Oracle and/or its affiliates. All rights reserved.
```
jonathan-gibbons, line 2: Year should probably be 2024

test/excludeverify/ExcludeVerifyTest.gmk
```diff
@@ -0,0 +1,149 @@
...
+		-w:$(@:%.ok=%/work) \
+		-r:$(@:%.ok=%/report) \
+		-exclude:$(TESTDIR)/excludeverify/ProblemList/format.1.txt \
+		-verifyexclude \
+		$(TESTDIR)/excludeverify \
+		> $(@:%.ok=%/log 2>&1) \
+	    && (cat $(@:%.ok=%/log) ; exit 1) \
+		|| (exit 0)
+	$(GREP) -s "Must follow:" $(@:%.ok=%/log 2>&1)
+	$(GREP) -s "The fully qualified test must exists." $(@:%.ok=%/log 2>&1) && exit 1 || exit 0
```
jonathan-gibbons, line 45: Here and on similar lines, the appearance of `2>&1` _inside_ the `$(@...)` expression seems weird to the point of wrong.  I don't know what you're trying to do of if any redirection is actually necessary here. If you copied this code from elsewhere, I'd be interested to know where from.

LudwikJaniuk:
@sormuras it's an interesting suggestion, I'll read up on --check.

LudwikJaniuk:
@sormuras So, having looked at -check: it "DOES NOT" run the tests. But I intend problemlist verification to be run as part of testing, e.g. in CI's. So I think this warrants existing outside of -check.

I agree -check should perhaps be extented to cover problemlists too, but that is not the scope of this PR.

Including -match seems like a very reasonable thing to add to this PR.

sormuras:
> I agree `-check` should perhaps be extented to cover `problemlists` too, but that is not the scope of this PR.

Okay. Let's introduce new options, including the "--long-word" variant: `--verify-exclude` in this PR.

test/excludeverify/ExcludeVerifyTest.gmk
LudwikJaniuk: I'm pretty sure I copied from https://github.com/openjdk/jtreg/blob/master/test/exclude/ExcludeTest.gmk, the same pattern is found on https://github.com/openjdk/jtreg/blob/master/test/exclude/ExcludeTest.gmk#L44.

test/excludeverify/ExcludeVerifyTest.gmk
```diff
@@ -0,0 +1,149 @@
...
+	$(GREP) -s "Exclude file cannot contain duplicate entries." $(@:%.ok=%/log 2>&1)
+	echo $@ passed at `date` > $@
+
+TESTS.jtreg += \
+	$(BUILDTESTDIR)/excludeverify.duplicate.ok \
+	$(BUILDTESTDIR)/excludeverify.format.1.ok \
+	$(BUILDTESTDIR)/excludeverify.format.2.ok \
+	$(BUILDTESTDIR)/excludeverify.good.ok \
+	$(BUILDTESTDIR)/excludeverify.id.ok \
+	$(BUILDTESTDIR)/excludeverify.exist.ok
```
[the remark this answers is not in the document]
LudwikJaniuk: Thank you.

LudwikJaniuk:
Ok so just after posting my comment about `-verifyexclude` vs `--verify-exclude`, I saw @sormuras comment 

> Let's introduce new options, including the "--long-word" variant: --verify-exclude in this PR.

I'm fine with this, but it's not what one infers from the current set of flags. So could we then document the intended style of new flags somewhere?

LudwikJaniuk:
I made `getMatchLists()` public in the last commit, just pointing that out, but I didn't see any reason it should not be.

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
jonathan-gibbons: The existing options are all "very old" and the style goes all the way back to the beginning of java-time, in the mid 90s. We have started using the double-dash form for recent options, especially where like-named options appear in other tools -- most notable `--class-path` etc.

When we started introducing the double-dash for in JDK 9 ([JEP 293](https://openjdk.org/jeps/293)) there was general agreement that we should not go back and "gratuitously" update existing old-form options. But equally, there was general agreement to use the new double-dash form for new options.

jonathan-gibbons:
> Ok so just after posting my comment about `-verifyexclude` vs `--verify-exclude`, I saw @sormuras comment
> 
> > Let's introduce new options, including the "--long-word" variant: --verify-exclude in this PR.
> 
> I'm fine with this, but it's not what one infers from the current set of flags. So could we then document the intended style of new flags somewhere?

[JEP 293](https://openjdk.org/jeps/293)

test/excludeverify/ExcludeVerifyTest.gmk
LudwikJaniuk: I am still not sure whether there's something wrong with these or not, but if there is, that other test should also be looked at.

LudwikJaniuk:
I've switched over to `--verify-exclude` / `-i` now. `-v`, `-e`, `-r` were either taken or felt misleading (verbose, exclude, recursive).

sormuras:
I'd omit the short form (flag) then. Having an expressive (almost self-explanatory) long form for "Verify Problemlist contents" is good.

test/excludeverify/ExcludeVerifyTest.gmk
jonathan-gibbons: The code might work "by accident", but is for sure not how it is meant to be.

There are two parts to this expression:
1. `$(@:%.ok=%/log)`. This is an expression to generate a string from `$@` by replacing the trailing `.ok` with `/log` -- in other words, based on the name of the target `$@` generate a related name for the log file
2. `2>&1`. This is plain old shell redirection, and should be directly on the `grep` command line, and not hidden in. the `$(@:...)` expression.

LudwikJaniuk: I've created https://bugs.openjdk.org/browse/CODETOOLS-7903699 to track the other occurrences.

LudwikJaniuk:
I believe I've addressed all comments. Is there anything remaining to be done before this can be approved?

src/share/classes/com/sun/javatest/regtest/tool/ExcludeFileVerifier.java
```diff
@@ -0,0 +1,150 @@
...
+        this.out = out;
+    }
+
+    private boolean hadErrors = false;
+    public boolean getHadErrors() {
+        return hadErrors;
+    }
+    public boolean verify(File file, List<String> validTestNames) {
+        ArrayList<String> usedTestNames = new ArrayList<String>();
+        ArrayList<Check> checks = new ArrayList<Check>();
```
jonathan-gibbons, line 53: If old school, it would be more appropriate to use just `List` on the LHS, if that is sufficient.

If new school, use `var`.

Either way, you don't need duplicate type parameters. Either use diamond syntax or `ver`.

src/share/classes/com/sun/javatest/regtest/tool/ExcludeFileVerifier.java
```diff
@@ -0,0 +1,150 @@
...
+        checks.add(new LineFormatCheck());
+        checks.add(new TestExistsCheck(validTestNames));
+        checks.add(new DuplicateCheck(usedTestNames));
```
jonathan-gibbons, line 56: I like the style of a potentially extensible set of checks ;-)

src/share/classes/com/sun/javatest/regtest/tool/ExcludeFileVerifier.java
```diff
@@ -0,0 +1,150 @@
...
+            System.out.println("File cannot be read: "  + file.getAbsolutePath());
+        }
+        return true;
+    }
+
+    static boolean lineIsComment(String line) {
+        line = line.trim();
+        if (line.equals("")) return true;
+        if (line.charAt(0) == '#') return true;
+        return false;
```
jonathan-gibbons, line 93: Could be simplified to 
```
return line.isBlank() || line.trim().startsWith("#");
```

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
```diff
@@ -1407,6 +1419,35 @@ public int run() throws BadArgs, Fault, Harness.Fault, InterruptedException {
...
+        }
+        boolean hadErrors = false;
+        for(File plf : excludeFiles) {
+            ExcludeFileVerifier efv = new ExcludeFileVerifier(out);
+            efv.verify(plf, validTestNames);
+            hadErrors |= efv.getHadErrors();
+        }
+
+        if (hadErrors) {
+            error("Cannot run because an exclude list had errors, printed above. Either resolve them or remove the exlude list.");
```
jonathan-gibbons, line 1447: typo: exlude

src/share/classes/com/sun/javatest/regtest/tool/i18n.properties
```diff
@@ -259,6 +259,7 @@ help.main.lock.desc=Lock file to use for tests in "exclusive access" directories
 help.main.nativepath.arg=<path>
 help.main.nativepath.desc=Path to location of native libraries and programs needed \
     by the tests.
+help.main.ve.desc=Verify contents of exclude files
```
jonathan-gibbons, line 262: "validate" may be a better word than "verify"

test/verifyexclude/VerifyExcludeTest.gmk
```diff
@@ -0,0 +1,190 @@
...
+
+TESTS.jtreg += \
+	$(BUILDTESTDIR)/verifyexclude.duplicate.ok \
+	$(BUILDTESTDIR)/verifyexclude.format.1.ok \
+	$(BUILDTESTDIR)/verifyexclude.format.2.ok \
+	$(BUILDTESTDIR)/verifyexclude.good.ok \
+	$(BUILDTESTDIR)/verifyexclude.id.ok \
+	$(BUILDTESTDIR)/verifyexclude.exist.ok \
+    $(BUILDTESTDIR)/verifyexclude.match.format.1.ok \
+    $(BUILDTESTDIR)/verifyexclude.shortform.ok
```
jonathan-gibbons, line 190: Something may have gone wrong with the indentation here

src/share/classes/com/sun/javatest/regtest/tool/ExcludeFileVerifier.java
LudwikJaniuk: Thank you!

src/share/classes/com/sun/javatest/regtest/tool/ExcludeFileVerifier.java
LudwikJaniuk: done

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
LudwikJaniuk: done

test/verifyexclude/VerifyExcludeTest.gmk
LudwikJaniuk: fixed

src/share/classes/com/sun/javatest/regtest/tool/i18n.properties
LudwikJaniuk: done

jonathan-gibbons (approved):

sormuras (changes requested):
Something is off with the `test/i18n/checkI18NProps.sh` test. Can you please take a look at those keys?

```
messages required but not defined:
help.main.i.desc

messages defined but not required:
help.main.ve.desc
```

https://github.com/LudwikJaniuk/jtreg/actions/runs/8465062864/job/23190939662#step:5:398

LudwikJaniuk:
Ah - taking a look

sormuras (approved):
All green now. Looks good!

PR #195, opened 2024-05-20
7903730: Enhance the agentvm to be tolerant to unexpected connection on the port the Agent listens on for handshake with the AgentServer

jaikiran:
Can I please get a review of this change which proposes to implement the enhancement noted in https://bugs.openjdk.org/browse/CODETOOLS-7903730?

As noted in that enhancement request, in the agentvm mode, we establish a socket connection between the `Agent` and the `AgentServer`. In its current form, it can so happen that the `Agent` may end up expecting a connection from some unexpected process. That then leads to the agent vm launch failing and the test too failing.

The proposed change introduces a handshake where the `AgentServer` sends the `jtreg` "magic" bytes immediately after it connects to the `Agent`. The `Agent` verifies these bytes on the accepted connection and if it either doesn't receive these bytes (for the specified read timeout) or if it receives some other data than these handshake bytes, then it closes this accepted connection and then goes on to accept another connection and repeats the handshake verification. In the current proposed change, it reattempts only 2 times and I haven't seen a need to either configure this value or have a higher value for this. i.e. if we can't establish a connection and verify that the connection initiator is the `AgentServer` even after three attempts, then I think it's OK to fail the agent launch process.

tier1 through tier8 tests have been run with this change (a couple of times) over the past couple of months and I haven't noticed any regression, nor have I noticed the agentvm launch process to be failing due to unexpected connection initiator.

---------
### Progress

### Issue
 * [CODETOOLS-7903730](https://bugs.openjdk.org/browse/CODETOOLS-7903730): Enhance the agentvm to be tolerant to unexpected connection on the port the Agent listens on for handshake with the AgentServer (**Enhancement** - P4)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/195/head:pull/195` \
`$ git checkout pull/195`

`$ git checkout pull/195` \
`$ git pull https://git.openjdk.org/jtreg.git pull/195/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 195`

`$ git pr show -t 195`

</details>

<a href="https://git.openjdk.org/jtreg/pull/195.diff">https://git.openjdk.org/jtreg/pull/195.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/195#issuecomment-2120484765)


LudwikJaniuk:
I'm excited for this!

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
```diff
@@ -182,12 +187,24 @@ private Agent(File dir, JDK jdk, List<String> vmOpts, Map<String, String> envVar
...
-                log("Waiting up to " + ACCEPT_TIMEOUT + " milli seconds for a" +
-                        " socket connection on port " + port +
-                        (pid != UNKNOWN_PID ? " from process " + pid : ""));
-                Socket s = ss.accept();
-                log("Received connection on port " + port + " from " + s);
-                s.setSoTimeout((int)(KeepAlive.READ_TIMEOUT * timeoutFactor));
+                final int readTimeout = (int) (KeepAlive.READ_TIMEOUT * timeoutFactor);
+                Socket s = null;
+                for (int i = 0; i < maxAcceptAttempts; i++) {
+                    log("Waiting up to " + ACCEPT_TIMEOUT + " milli seconds for a" +
```
LudwikJaniuk, line 193: Typo: milliseconds.

LudwikJaniuk: Oh, nevermind, it's not new.

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
```diff
@@ -212,6 +229,46 @@ private Agent(File dir, JDK jdk, List<String> vmOpts, Map<String, String> envVar
         }
     }
 
+    // accept()s a connection over the ServerSocket and then reads handshake bytes
+    // from the connected socket. if the read bytes match the expected handshake bytes
+    // then the connection and handshake is considered successful and the accepted
+    // Socket is returned. if the read bytes don't match the expected handshake bytes
+    // or if the read times out after accept()ing an connection, then this method
+    // closes the accepted connection and returns null.
```
LudwikJaniuk, line 237: Please capitalize the comments.

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
jaikiran: Hello Ludvig, I think it was me who introduced that log message in the first place (in some previous integrated code). I've fixed it now and updated the PR.

src/share/classes/com/sun/javatest/regtest/agent/AgentServer.java
```diff
@@ -231,6 +242,12 @@ public void close() {
         }
     }
 
+    private static void writeHandshakeMagicBytes(final Socket s) throws IOException {
+        final OutputStream os = s.getOutputStream();
+        os.write(JTREG_AGENT_HANDSHAKE_MAGIC);
+        os.flush();
```
LudwikJaniuk, line 248: Is this the point where the agent would break the connection if the magic bytes are wrong? I'm trying to more clearly see at which point the jtreg "server" knows that the jtreg agent has accepted the handshake.

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
```diff
@@ -212,6 +229,46 @@ private Agent(File dir, JDK jdk, List<String> vmOpts, Map<String, String> envVar
...
+                        + Arrays.toString(JTREG_AGENT_HANDSHAKE_MAGIC)
+                        + ", actual: " + Arrays.toString(handshakeBytes));
+                return null;
+            }
+        } catch (IOException ioe) {
+            log("closing connection to " + s + " due to: " + ioe);
+            try {
+                s.close();
+            } catch (IOException ignored) {
+                // ignore
```
LudwikJaniuk, line 264: Please don't swallow exceptions.

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
jaikiran: Fixed.

src/share/classes/com/sun/javatest/regtest/agent/AgentServer.java
jaikiran: > Is this the point where the agent would break the connection if the magic bytes are wrong?

That happens in the `Agent` code in the newly introduced `acceptAndHandshake()` method  https://github.com/openjdk/jtreg/pull/195/files#diff-23190c9a688c97b3edf1b5605b78eefb8ebe4b5894737518072c6efb55705401R252. The `Agent` is the server side of the socket connection, listening for a connection, and the `AgentServer` is the client side of the socket connection sending a connection request. Yes, it's a bit confusing since the class is named `AgentServer`, but it does act as a server later on, serving test execution requests from the jtreg's `Agent`.

dfuch:
Not a reviewer here, but it looks reasonable to me. You could possibly have used `InputStream::readNBytes` to simplify the code.

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
jonathan-gibbons: Use `Throwable.addSuppressed` -- it is for exactly this situation.

jonathan-gibbons (commented):
Is there a reason to use a static (fixed) string for the handshake, as compared to a dynamically generated string?

jaikiran:
Hello Daniel,

> You could possibly have used `InputStream::readNBytes` to simplify the code. Unless you're constrained to use JDK 7/8 APIs?

I did attempt to simplify that part to use `InputStream.readNBytes()` but like you note, jtreg is constrainted to Java 8 APIs.

jaikiran:
Hello Jon,

> Is there a reason to use a static (fixed) string for the handshake, as compared to a dynamically generated string?

My first version which I experimented with actually used a 1 byte random content as the handshake data (literally used `java.util.Random` and did a bit more by making sure the value was above a certain floor value). The random content would be generated afresh every time a new `AgentServer` is being launched by the `Agent`. The `Agent` when launching the `AgentServer` would pass this data as a new command line argument `--handshake 0xXX`. The `AgentServer` would then, upon connecting, send this dynamic handshake data to the `Agent` and the `Agent` would verify that it matches the expected data.

That approach worked fine too. But when thinking about the goal, I realized that we aren't trying to prevent processes which are intentionally trying to interfere with jtreg by connecting to the jtreg port. Instead our goal is to prevent unintentional interference. So I thought having a well known, static handshake byte data (`jtreg`) that applications might know about, should be OK. You wouldn't expect them to intentionally send out this handshake data to interfere with jtreg. So I decided to keep it simple and use the static data.

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
jaikiran: > Use Throwable.addSuppressed -- it is for exactly this situation.

But we wouldn't be throwing any exception to which we would be adding this as a suppressed exception. This failure is during close()ing a connection on which we either had a read timeout or an IOException when reading off it. We log that failure a few lines above this and then go and close this connection and then attempt to accept() a next connection (if any).

I can update this catch block to log a message which states the close() failed, along with the reason. Is that reasonable?

src/share/classes/com/sun/javatest/regtest/agent/AgentServer.java
LudwikJaniuk: Right, that's confusing but thank you for explaining. 
When reading "handshake" I expected both sides to send each other some indication of having accepted the connection. Compare with e.g. the TCP three-way handshare https://en.wikipedia.org/wiki/Handshake_(computing)#TCP_three-way_handshake. This protocol seems to be a "one-way handshake" in that, the `AgentServer` acting as client connects and identifies itself with some magic bytes. The `Agent` acting as server can decline if the bytes don't match. The `AgentServer` however must simply assume that its magic bytes were accepted. Have I gotten that right? And if so, is the term "handshake" really appropriate or would a term such as "connection identification" or something else be more fitting?

src/share/classes/com/sun/javatest/regtest/agent/AgentServer.java
```diff
@@ -210,6 +218,9 @@ public void close() {
             log("Connecting to " + host + ":" + port);
             Socket s = new Socket(host, port);
```
msheppar, line 219: This only works if the host argument is not specified on the AgentServer command line, as the Agent now listens on the loopback address  and no longer the wildcard address.
Typically the host argument, in the launching command line of the AgentServer, is not specified. But now an Agent  is listening on loopback address making the -host argument redundant. So maybe -host should be removed from common line processing

src/share/classes/com/sun/javatest/regtest/agent/AgentServer.java
```diff
@@ -55,10 +55,18 @@
...
+
 
 @SuppressWarnings("removal") // Security Manager and related APIs
 public final class AgentServer implements ActionHelper.OutputHandler {
 
+    /**
+     * The handshake bytes that the AgentServer sends over a Socket to the Agent
+     * to complete the communication handshake
+     */
+    public static final byte[] JTREG_AGENT_HANDSHAKE_MAGIC = "jtreg".getBytes(US_ASCII);
```
msheppar, line 68: I was developing a similar PoC for AgentSevre to Agent connectivity
I used the pattern "Hello Agent:"<id>
using the Agent is in the connection string, which the Agent compared with its own id as a condition of "accepting" the AgentServer connection

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
```diff
@@ -147,7 +148,11 @@ private Agent(File dir, JDK jdk, List<String> vmOpts, Map<String, String> envVar
             // is platform-specific, and Solaris has it on by default.
             ss.setReuseAddress(false);
             InetAddress loopbackAddr = InetAddress.getLoopbackAddress();
-            ss.bind(new InetSocketAddress(loopbackAddr, /*port:*/ 0), /*backlog:*/ 1);
+            // We allow for some unexpected processes to connect to the port we listen on.
+            // We discard such connections if, after connecting, our internal handshake fails.
+            final int maxAcceptAttempts = 3;
```
msheppar, line 153: this is effectively a constant -- maybe have it as a private static final MAXACCEPTATTEMPTS = 3;

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
```diff
@@ -212,6 +229,59 @@ private Agent(File dir, JDK jdk, List<String> vmOpts, Map<String, String> envVar
...
+            log("closing connection to " + s + " due to: " + ioe);
+            try {
+                s.close();
+            } catch (IOException ignored) {
+                // ignore
+            }
+            return null;
+        }
+        // we don't read more than handshake bytes in this method
+        assert totalRead <= JTREG_AGENT_HANDSHAKE_MAGIC.length : "unexpected number of handshake" +
```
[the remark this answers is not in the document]
dfuch: It would be an error if totalRead was > JTREG_AGENT_MAGIC.length: according to the logic of the code that reads the magic that cannot happen, ever. Hence the assert.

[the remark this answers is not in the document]
msheppar: JP, yes understood
I should have been more direct and precise in my comment i.e. 
the assert is not needed — restructure the if statement.

The condition can’t occur and if it did occur then what would that mean?
Is there a problem with either the code generated by Java ? or 
that the java runtime has a problem? or the return from InputStream.read is greater than expected, 
thus the InputStream has a problem?

An AgentServer is (nearly) always launched with “-ea” option — only if jtreg is launched from command line without -ea will it not have asserts enabled.

Looking at the current logic, rather than testing all negative cases, it can be restructured with

if (totalRead != JTREG_AGENT_HANDSHAKE_MAGIC.length) {
   // problem
} else {
   // check for connect message
}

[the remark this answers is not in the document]
dfuch: So we could do:
```
if (totalRead != JTREG_AGENT_HANDSHAKE_MAGIC.length) {
    assert totalRead < JTREG_AGENT_HANDSHAKE_MAGIC.length
   // problem
} else {
   // check for connect message
}
```

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
jaikiran: Done. I've now moved this to a `private static final` field.

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
jaikiran: I've now updated the PR to log the exception stacktrace if/when it fails. This should help debug any unexpected failures in this part of the code.

src/share/classes/com/sun/javatest/regtest/exec/Agent.java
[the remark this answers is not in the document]
jaikiran: I agree with Daniel about the useful of `assert`. I have updated the PR to use Daniel's suggested approach in that snippet.

jaikiran:
Thank you everyone for the review comments and inputs so far and sorry for not responding to this earlier. I wanted to finish off some JDK mainline work before rampdown, so I had kept this jtreg PR on the side for a few days. I have now refreshed this with the suggested changes. Existing self tests continue to pass with the current updates. I plan to trigger an extensive CI run in a day or two.

src/share/classes/com/sun/javatest/regtest/agent/AgentServer.java
jaikiran: Hello Mark, I read this a couple of times but didn't fully understand this suggestion. Right now the `host` variable gets default set to loopback address:

```
// by default use loopback address. if "-host" is specified, this will
// then be overridden by that value
InetAddress host = InetAddress.getLoopbackAddress();

```
but as the code comment states, if the `AgentServer` program was launched with a `-host <val>` then the `host` variable here is updated/set to the new value that is provided to value. So when the code does indeed reach the point of calling this `Socket` constructor, the `host` variable would have the correct host.
Did I misunderstand your suggestion?

src/share/classes/com/sun/javatest/regtest/agent/AgentServer.java
jaikiran: I had given that a thought too - the random handshake bytes that I note in a previous comment of mine https://github.com/openjdk/jtreg/pull/195#issuecomment-2121551694 is a variant of this approach where each `AgentServer` process handshakes with the `Agent` using a different handshake data. For reasons I state in that other comment, I felt that using a dynamic handshake value isn't adding much value and so decided to use a static handshake value instead.

msheppar: of course you can accept or reject a suggestion, but it does add value in that it pairs the agent server and the agent by verifying the agent id passed to the agent server

src/share/classes/com/sun/javatest/regtest/agent/AgentServer.java
msheppar: your recent change to the Agent to use the loopback address only

 ServerSocket ss = new ServerSocket();
            // Ensure SO_REUSEADDR is false. (It is only needed if we're
            // using a fixed port.) The default setting for SO_REUSEADDR
            // is platform-specific, and Solaris has it on by default.
            ss.setReuseAddress(false);
            InetAddress loopbackAddr = InetAddress.getLoopbackAddress();
            final int backlog = MAX_ACCEPT_ATTEMPTS;
            ss.bind(new InetSocketAddress(loopbackAddr, /*port:*/ 0), backlog);
            final int port = ss.getLocalPort();

so the -host argument is now redundant, if it were passed as an argument then an AgentServer couldn't connect to its peer Agent

jaikiran:
Keep open, I'm experimenting with some suggested changes before finalizing this PR in the coming days.

PR #205, opened 2024-06-10
7902346: Need better message when test times out in agentvm mode

jaikiran:
Can I please get a review of this change which proposes to improve the error message when a test action times out on an agent VM? This addresses https://bugs.openjdk.org/browse/CODETOOLS-7902346.

Before this change, whenever a test action timed out in an agentvm, then it would be reported as:

> result: Error. Agent error: java.lang.Exception: Agent 3 timed out with a timeout of 1200 seconds; check console log for any additional details 

With the changes in this PR, the error reporting will be a more accurate and won't give an impression that this was some error within the jtreg agent implementation. For `@run main ...`, the error will now say:

> result: Error. "main" action timed out with a timeout of 1200 seconds on agent 3

Another example, this time for a driver action:

> result: Error. "driver" action timed out with a timeout of 1200 seconds on agent 3

A couple of new self tests have been added to verify this change. Self tests including this new one continue to pass with this change.

---------
### Progress

### Issue
 * [CODETOOLS-7902346](https://bugs.openjdk.org/browse/CODETOOLS-7902346): Need better message when test times out in agentvm mode (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/205/head:pull/205` \
`$ git checkout pull/205`

`$ git checkout pull/205` \
`$ git pull https://git.openjdk.org/jtreg.git pull/205/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 205`

`$ git pr show -t 205`

</details>

<a href="https://git.openjdk.org/jtreg/pull/205.diff">https://git.openjdk.org/jtreg/pull/205.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/205#issuecomment-2157526410)

### DIFF
--- a/src/share/classes/com/sun/javatest/regtest/exec/Agent.java
+++ b/src/share/classes/com/sun/javatest/regtest/exec/Agent.java
@@ -87,6 +87,12 @@ public class Agent {
         }
     }
 
+    // represents a timeout that occurred while executing
+    // a test action within an agentvm
+    static final class ActionTimeout extends Exception {
+        private static final long serialVersionUID = 7956108605006221253L;
+    }
+
     // legacy support for logging to stderr
     // showAgent is superseded by always-on log to file
     static final boolean showAgent = Flags.get("showAgent");
@@ -309,7 +315,7 @@ public class Agent {
             int timeout,
             final TimeoutHandler timeoutHandler,
             TestResult.Section trs)
-                throws Fault {
+                throws ActionTimeout, Fault {
         trace("doCompileAction " + testName + " " + cmdArgs);
 
         return doAction("doCompileAction",
@@ -342,7 +348,7 @@ public class Agent {
             int timeout,
             final TimeoutHandler timeoutHandler,
             TestResult.Section trs)
-                throws Fault {
+                throws ActionTimeout, Fault {
         trace("doMainAction: " + testName
                     + " " + testClassPath
                     + " " + modulePath
@@ -382,7 +388,7 @@ public class Agent {
             int timeout,
             final TimeoutHandler timeoutHandler,
             TestResult.Section trs)
-                throws Fault {
+                throws ActionTimeout, Fault {
         final PrintWriter messageWriter = trs.getMessageWriter();
         // Handle the timeout here (instead of in the agent) to make it possible
         // to see the unchanged state of the Agent JVM when the timeout happens.
@@ -417,8 +423,7 @@ public class Agent {
             keepAlive.setEnabled(true);
             if (alarm.didFire()) {
                 waitForTimeoutHandler(actionName, timeoutHandler, timeoutHandlerDone);
-                throw new Fault(new Exception("Agent " + id + " timed out with a timeout of "
-                        + timeout + " seconds"));
+                throw new ActionTimeout();
             }
         }
     }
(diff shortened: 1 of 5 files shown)


msheppar:
Here's my tuppence worth.
I think the existing  message is reasonable clear as it indicates that the test being executed by the agentvm i.e. the AgenetServer, has failed to complete in the expected jtreg time frame
The newer message add additional ambiguity as it is not the main action that causes the timeout but some problem within the test e.g. a UDP socket receive or a test thread that has not completed and been reaped successfully.

I think amending the exist message with a little guidance would be preferable, for example,

result: Error. Agent error: java.lang.Exception: Agent 3 timed out with a timeout of 1200 seconds; check console log for any additional details and process capture in process.html

jonathan-gibbons (approved):
LGTM

msheppar:
As an extensive user of this error and its implications, I have to disagree with Jonathon on this. I don't think that the additions in their current form assist in the diagnosis of a failure. They are even more obscure than the Agent error message It is not a jtreg failure per se, it a failure of a test to execute in a timely manner. 

Providing an error or exception message that indicates main action or driver action doesn't help in the diagnosis. 

Knowing the Agent/AgentServer relationship and conveying that a timeout on test completion has occurred is important. Also, providing guidance on where to look for further details i.e. to look in process capture (process.html), would be helpful

jaikiran:
Hello Mark,

> result: Error. Agent error: java.lang.Exception: Agent 3 timed out with a timeout of 1200 seconds; check console log for any additional details and process capture in process.html

The processes.html isn't a construct of jtreg itself. jtreg allows for failure/timeout handlers to be invoked upon timeout. The JDK project (which is just one user of jtreg) is configured to run a failure handler to generate the processes.html. So jtreg itself can write out this message pointing users to that processes.html.

jaikiran:
> I think amending the exist message with a little guidance would be preferable, for example,

> result: Error. Agent error: java.lang.Exception: 

The original improvement proposed in this enhancement is to make it clear that the failure isn't an "Agent error". Instead it is a timeout (that can be caused due to several reasons) of an action (the one which we now report) that was running on the agent.

msheppar:
> Hello Mark,
> 
> > result: Error. Agent error: java.lang.Exception: Agent 3 timed out with a timeout of 1200 seconds; check console log for any additional details and process capture in process.html
> 
> The processes.html isn't a construct of jtreg itself. jtreg allows for failure/timeout handlers to be invoked upon timeout. The JDK project (which is just one user of jtreg) is configured to run a failure handler to generate the processes.html. So jtreg itself cannot write out this message pointing users to that processes.html.

ok ... yes, that's reasonable point ... if the build hasn't been configured with sh jib.sh configure -- --enable-jtreg-failure-handler then the failure handler won't be run

msheppar:
> > I think amending the exist message with a little guidance would be preferable, for example,
> 
> > result: Error. Agent error: java.lang.Exception:
> 
> The original improvement proposed in this enhancement is to make it clear that the failure isn't an "Agent error". Instead it is a timeout (that can be caused due to several reasons) of an action (the one which we now report) that was running on the agent.

But the action is within the Agent ;-)   so Agent Error is not inappropriate
My preference is a refinement of the current message
But, in any case the change has been approved, so let'd not over-egg the issue

a ubiquitous LGTM

jaikiran:
Thank you Mark and Jon for your inputs and reviews.

PR #208, opened 2024-06-19
7903188: Log time spent waiting to acquire exclusive access lock

jaikiran:
Can I please get a review of this change which proposes to implement the enhancement requested in https://bugs.openjdk.org/browse/CODETOOLS-7903188?

jtreg supports the ability to sequentially execute tests, instead of concurrently, for tests belonging to a pre-configured `exclusiveAccess.dirs` directory. The `MainAction` and `ShellAction` before launching the test, first acquire a lock. The lock acquisition can be time consuming and depends on how long an already running test from that directory takes to complete. This lock acquisiton time isn't reported anywhere in the jtreg action's `section` in the report. Because of this, it sometimes makes it difficult to determine where the unaccounted time is spent.

The change in this PR prints out how long it took to acquire a exclusive access before launching the test. It's only printed if the test was configured with exclusiveAccess. The reported message will look like:

```
#section:main
----------messages:(7/230)----------
command: main Test
reason: User specified action: run main Test 
started: Wed Jun 19 14:45:21 IST 2024
exclusiveAccess wait time (seconds): 20.289
Mode: othervm
finished: Wed Jun 19 14:45:21 IST 2024
elapsed time (seconds): 0.123
```

An existing self test has been updated to verify this change. All existing tests continue to pass with this change.

### DIFF
--- a/src/share/classes/com/sun/javatest/regtest/exec/Action.java
+++ b/src/share/classes/com/sun/javatest/regtest/exec/Action.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -711,6 +711,7 @@ public abstract class Action extends ActionHelper {
         LOG_JT_COMMAND        = "JavaTest command: ",
         LOG_REASON            = "reason: ",
         LOG_ELAPSED_TIME      = "elapsed time (seconds): ",
+        LOG_EXCLUSIVE_ACCESS_TIME = "exclusiveAccess wait time (seconds): ",
         LOG_STARTED = "started: ",
         LOG_FINISHED = "finished: ",
         //LOG_JDK               = "JDK under test: ",
--- a/src/share/classes/com/sun/javatest/regtest/exec/MainAction.java
+++ b/src/share/classes/com/sun/javatest/regtest/exec/MainAction.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2023, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -31,6 +31,7 @@ import java.io.FileWriter;
 import java.io.IOException;
 import java.io.PrintWriter;
 import java.nio.file.Path;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.EnumSet;
@@ -338,11 +339,20 @@ public class MainAction extends Action
             endAction(status);
         } else {
             Lock lock = script.getLockIfRequired();
-            if (lock != null) lock.lock();
-
+            long exclusiveAccessWaitMillis = 0;
+            if (lock != null) {
+                long startNanos = System.nanoTime();
+                lock.lock();
+                exclusiveAccessWaitMillis = Duration.ofNanos(System.nanoTime() - startNanos).toMillis();
+            }
             // Start action after the lock is taken to ensure correct "elapsed time".
             startAction(true);
-
+            if (lock != null) {
+                // print the duration we waited for acquiring a lock due to
+                // exclusiveAccess.dir
+                section.getMessageWriter().println(LOG_EXCLUSIVE_ACCESS_TIME
+                        + ((double) exclusiveAccessWaitMillis / 1000.0));
+            }
             try {
                 switch (!othervmOverrideReasons.isEmpty() ? ExecMode.OTHERVM : script.getExecMode()) {
                     case AGENTVM:
--- a/src/share/classes/com/sun/javatest/regtest/exec/ShellAction.java
+++ b/src/share/classes/com/sun/javatest/regtest/exec/ShellAction.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1998, 2022, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1998, 2024, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -29,6 +29,7 @@ import java.io.File;
 import java.io.PrintWriter;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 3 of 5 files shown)


AlanBateman:
```
#section:main
----------messages:(7/230)----------
command: main Test
reason: User specified action: run main Test 
started: Wed Jun 19 14:45:21 IST 2024
exclusiveAccess wait time (seconds): 20.289
Mode: othervm
finished: Wed Jun 19 14:45:21 IST 2024
elapsed time (seconds): 0.123
```

I wonder if "exclusiveAccess wait time ..." could be printed before the "started ..." line. That is, the waiting happens before the test starts so would be confusing to have a "wait" log message between "started" and "finished".

jaikiran:
Hello Alan,

> I wonder if "exclusiveAccess wait time ..." could be printed before the "started ..." line. That is, the waiting happens before the test starts so would be confusing to have a "wait" log message between "started" and "finished".

I had thought of that, but it would need a slightly more involved change because the construct that allows the code write these messages (the `section`) is instantiated within the `startAction()` method which writes the `started` message. So it's only after that `startAction()` method returns, we get a `section` instance through which we can write out the exclusive access wait time. 

Having said that, I do agree with what you state. Let me see if I can keep it simple and yet be able to write it in that sequence.

jaikiran:
I've updated the PR to clean up the logic of lock acquisition, printing the wait time and the lock release. The updated change centralizes this responsibility into the base `com.sun.javatest.regtest.exec.Action` class which simplifies dealing with the lock as well as printing the wait time.

With these changes, the output will now look like:

```
#section:main
----------messages:(7/229)----------
command: main Test
reason: User specified action: run main Test 
exclusiveAccess wait time (seconds): 20.304
started: Thu Jun 20 15:55:38 IST 2024
Mode: othervm
finished: Thu Jun 20 15:55:38 IST 2024
elapsed time (seconds): 0.13
```

All tests continue to pass.

src/share/classes/com/sun/javatest/regtest/config/RegressionTestSuite.java
```diff
@@ -215,10 +215,26 @@ public boolean useOtherVM(TestDescription td) throws TestSuite.Fault {
...
 
+    /**
+     * @param td the test description
+     * {@return true if the test is configured to run exclusively, false otherwise}
+     * @deprecated Use {@link #isExclusiveAccessRequired(TestDescription)} instead,
+     *             which doesn't throw the checked exception
+     * @throws TestSuite.Fault
+     */
+    @Deprecated
     public boolean needsExclusiveAccess(TestDescription td) throws TestSuite.Fault {
```
jaikiran, line 226: This already existing `throws TestSuite.Fault` is actually unnecessary because nothing in that logic throws that checked exception. However, I decided not to change this method signature since that can cause compatibility issue - I'm guessing this `com.sun.javatest.regtest.config.RegressionTestSuite` is a public API?
That's why I decided to introduce a new method which doesn't throw the checked exception.

sormuras: I am not aware of any consumer of `RegressionTestSuite` - is there, @jonathan-gibbons ?

sormuras (approved):
These changes look correct and cleaner than before.

src/share/classes/com/sun/javatest/regtest/config/RegressionTestSuite.java
```diff
@@ -215,10 +215,26 @@ public boolean useOtherVM(TestDescription td) throws TestSuite.Fault {
         return properties.useOtherVM(td.getFile());
     }
 
+    /**
+     * @param td the test description
+     * {@return true if the test is configured to run exclusively, false otherwise}
```
jonathan-gibbons, line 220: This is in the wrong place.
As written, the inline tag `{@return...}` is part of the description of the preceding block tag `@param`

When using the inline form of the `return` tag, it should appear at the beginning of the description.

src/share/classes/com/sun/javatest/regtest/config/RegressionTestSuite.java
jonathan-gibbons: `RegressionTestSuite` is a mostly-private implementation of a public interface `TestSuite`.
"mostly private" means that there are probably no other uses of the class. Maybe at one point I thought there might be.

I also suspect this might be a historical leftover from times past, when access to the underlying properties might have been more lazy.

At any rate, any clients of `RegressionTestSuite` would be "friends" and should be willing to tolerate change. There is no published commitment to the stability of these internal classes.

src/share/classes/com/sun/javatest/regtest/config/RegressionTestSuite.java
```diff
@@ -215,10 +215,26 @@ public boolean useOtherVM(TestDescription td) throws TestSuite.Fault {
...
+     * @throws TestSuite.Fault
+     */
+    @Deprecated
     public boolean needsExclusiveAccess(TestDescription td) throws TestSuite.Fault {
         return properties.needsExclusiveAccess(td.getFile());
     }
 
+    /**
+     * @param td the test description
+     * {@return true if the test is configured to run exclusively, false otherwise}
```
jonathan-gibbons, line 232: same bad `{@return}` tag as above.

jonathan-gibbons (changes requested):
Some bad javadoc comments; not a show stopper, but worth addressing to reduce e future confusion.

src/share/classes/com/sun/javatest/regtest/config/RegressionTestSuite.java
jaikiran: > At any rate, any clients of RegressionTestSuite would be "friends" and should be willing to tolerate change. There is no published commitment to the stability of these internal classes.

That makes things simpler and cleaner too. I've now updated the PR to not introduce another method and instead just update the signature of the existing one.

src/share/classes/com/sun/javatest/regtest/config/RegressionTestSuite.java
jaikiran: Hello Jon, I wasn't aware of this. I've now moved the `{@return}` to place it before the `@param`.

jaikiran:
Tests continue to pass with the latest updated PR.

src/share/classes/com/sun/javatest/regtest/config/RegressionTestSuite.java
jonathan-gibbons: Generally, the format of a doc comment is an initial _description_ followed by _block tags_.  Block tags begin with `@` at the beginning of a line. Some (but not all) block tags may also contain descriptions, perhaps after some other token (e.g. `@param`  `@throws`).
Descriptions are typically composed of plain text, HTML elements, and _inline tags_. Inline tags are all of the form `{@`_name_ _optional-args_`}` -- and soon, depending on the doc comment, Markdown constructs.  Some inline tags (`{@return...}` and `{@summary ...}`) must appear at the beginning of the initial description.

For more info, see the Doc Comment Specification in the main JDK docs, such as [here](https://docs.oracle.com/en/java/javase/21/docs/specs/javadoc/doc-comment-spec.html)
You can generally always find this with either a Google search or a search in the JDK API search box for "Doc Comment Spec".

jonathan-gibbons: FWIW, the inline form `{@return `_description_`}` is just a short form for
`Returns` _description_ `.`
@return description

which is a common but repetitive idiom in doc comments.

jonathan-gibbons (approved):

src/share/classes/com/sun/javatest/regtest/config/RegressionTestSuite.java
jaikiran: Thank you Jon for those helpful details.

jaikiran:
Thank you Alan and Jon for the inputs and reviews.

PR #211, opened 2024-07-02
7903765: wget failed in build.sh in jtreg

YaSuenag:
I tried to build jtreg, but it failed as following:

```
$ bash make/build.sh --jdk /usr/lib/jvm/java-22-ope
njdk
[build.sh][INFO] CYGWIN_OR_MSYS=0
[build.sh][INFO] JAVA_HOME: /usr/lib/jvm/java-22-openjdk
[build.sh][INFO] Downloading https://archive.apache.org/dist/ant/binaries/apache-ant-1.10.8-bin.zip to /home/ysuenaga/github-forked/jtreg/make/../build/deps/apache-ant-1.10.8-bin.zip
[build.sh][ERROR] wget exited with exit code 1
```

I tried it on Fedora 40, wget is from wget2-wget-2.1.0-9.fc40.x86_64 . Related path seems not to be accepted in this version at least.

### DIFF
--- a/make/build-support/build-common.sh
+++ b/make/build-support/build-common.sh
@@ -106,7 +106,7 @@ download_using_wget() {
     check_arguments "${FUNCNAME}" 2 $#
 
     local url="$1"
-    local destfile="$2"
+    local destfile=`realpath "$2"`
 
     set +e
     "${WGET}" ${WGET_OPTIONS} "${url}" -O "${destfile}"


jaikiran:
Hello @YaSuenag, it's not clear why it is failing. Can you maybe run that wget command explicitly from the command line and see what error messages you see?

YaSuenag:
> Can you maybe run that wget command explicitly from the command line and see what error messages you see?

I saw following log when I run build.sh with `bash -x`

```
+ /usr/bin/wget -q https://archive.apache.org/dist/ant/binaries/apache-ant-1.10.8-bin.zip -O /home/ysuenaga/github/jtreg/make/../build/deps/apache-ant-1.10.8-bin.zip
+ ret=1
+ '[' '!' 1 = 0 ']'
+ error 'wget exited with exit code 1'
+ log_message ERROR 'wget exited with exit code 1'
```

When I run `wget` without `-q`, it reports `wget` failed  due to relative path.

```
$ wget https://archive.apache.org/dist/ant/binaries/apache-ant-1.10.8-bin.zip -O /home/ysuenaga/github/jtreg/make/../build/deps
/apache-ant-1.10.8-bin.zip
Internal error: Unexpected relative path: '/home/ysuenaga/github/jtreg/make/..'
```

jaikiran:
It's strange that wget is complaining for that path. Instead of using `realpath` at the call site, can you see the following patch works on your setup. What it does is, it avoids the use of `..` in paths that these scripts use:

```diff
diff --git a/make/build-support/asmtools/build.sh b/make/build-support/asmtools/build.sh
index 0d2762f..0662d56 100644
--- a/make/build-support/asmtools/build.sh
+++ b/make/build-support/asmtools/build.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 #
-# Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2020, 2024, Oracle and/or its affiliates. All rights reserved.
 # DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
 #
 # This code is free software; you can redistribute it and/or modify it
@@ -42,8 +42,8 @@ setup_asmtools_src() {
     check_arguments "${FUNCNAME}" 1 $#
 
     local dir="$1"
-
-    local ASMTOOLS_LOCAL_SRC_ARCHIVE="${dir}/../source.zip"
+    local src_archive_dir="$(builtin  cd ${dir}/..; pwd)"
+    local ASMTOOLS_LOCAL_SRC_ARCHIVE="${src_archive_dir}/source.zip"
     if [ "${ASMTOOLS_SRC_TAG}" = "tip" -o "${ASMTOOLS_SRC_TAG}" = "master" ]; then
         local BRANCH="master"
         get_archive_no_checksum "${CODE_TOOLS_URL_BASE}/asmtools/archive/${BRANCH}.zip" "${ASMTOOLS_LOCAL_SRC_ARCHIVE}" "${dir}"
@@ -58,6 +58,7 @@ build_asmtools() {
     check_arguments "${FUNCNAME}" 0 $#
 
     local ASMTOOLS_SRC_DIR_BASE="${BUILD_DIR}/src"
+    mkdir -p "${ASMTOOLS_SRC_DIR_BASE}"
     setup_asmtools_src "${ASMTOOLS_SRC_DIR_BASE}"
 
     local ASMTOOLS_DIST="${BUILD_DIR}/build"
diff --git a/make/build-support/build-common.sh b/make/build-support/build-common.sh
index e5603ea..938fe41 100644
--- a/make/build-support/build-common.sh
+++ b/make/build-support/build-common.sh
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2020, 2024, Oracle and/or its affiliates. All rights reserved.
 # DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
 #
 # This code is free software; you can redistribute it and/or modify it
@@ -311,8 +311,8 @@ if [ -z "${log_module:-}" ]; then
     error "log_module not set in caller (line/file): $(caller)"
     exit 1
 fi
-
-ROOT="$(abspath ${ROOT:-${mydir}/..})"
+DEFAULT_ROOT="$(builtin cd ${mydir}/..; pwd)"
+ROOT="$(abspath ${ROOT:-${DEFAULT_ROOT}})"
 BUILD_DIR="$(abspath "${BUILD_DIR:-${ROOT}/build}")"
 DEPS_DIR="${BUILD_DIR}/deps"
 
diff --git a/make/build-support/jtharness/build.sh b/make/build-support/jtharness/build.sh
index f76a823..62ca0d4 100644
--- a/make/build-support/jtharness/build.sh
+++ b/make/build-support/jtharness/build.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 #
-# Copyright (c) 2020, Oracle and/or its affiliates. All rights reserved.
+# Copyright (c) 2020, 2024, Oracle and/or its affiliates. All rights reserved.
 # DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
 #
 # This code is free software; you can redistribute it and/or modify it
@@ -46,7 +46,8 @@ setup_jtharness_source() {
     local dir="$1"
 
     # Build jtharness
-    local JTHARNESS_LOCAL_SRC_ARCHIVE="${dir}/../source.zip"
+    local src_archive_dir="$(builtin  cd ${dir}/..; pwd)"
+    local JTHARNESS_LOCAL_SRC_ARCHIVE="${src_archive_dir}/source.zip"
     if [ "${JTHARNESS_SRC_TAG}" = "tip" -o "${JTHARNESS_SRC_TAG}" = "master" ]; then
         local BRANCH="master"
         get_archive_no_checksum "${CODE_TOOLS_URL_BASE}/jtharness/archive/${BRANCH}.zip" "${JTHARNESS_LOCAL_SRC_ARCHIVE}" "${dir}"
@@ -61,6 +62,7 @@ build_jtharness() {
     check_arguments "${FUNCNAME}" 0 $#
 
     local JTHARNESS_SRC_DIR_BASE="${BUILD_DIR}/src"
+    mkdir -p "${JTHARNESS_SRC_DIR_BASE}"
     setup_jtharness_source "${JTHARNESS_SRC_DIR_BASE}"
 
     local JTHARNESS_DIST="${BUILD_DIR}/build"


```

jaikiran:
Thank you for verifying it. I will let Jon and others more familiar with these scripts decide which one would be better.

YaSuenag:
IMHO the patch should be simple if it has consistency.
Relative path (`..`) is valid in general and it is nature to use. So I think we can use `realpath` in this case.

Anyway, I'm waiting the conclusion of this issue. I can close this when another PR is sent.

jonathan-gibbons:
Catching up with this thread ...

Overall, while @jaikiran's solution is more complex, I think it is a better and broader solution.

While the original proposal is simpler, it is less obvious -- when I was reading the ibnitial patch (before seeing the whole thread here)  I was wondering why `realpath` was apparently necessary in this specific situation.  So while I agree that `..` is natural and should be OK to use, it is clear that at least one tool does not like the use of `..` for whatever reason it thinks that `..` is bad.  If nothing else, if we deem that `..` is bad, we should at least avoid it at the point at which it is being introduced into paths, and not at the point of use for tools that don't like it.

jonathan-gibbons:
FWIW, a Google search on that specific error message indicates this is likely a bug in that version of `wget`.
See here: https://lists.gnu.org/archive/html/wget-dev/2020-05/msg00013.html

jaikiran:
Thank you Jon for the inputs. 

Hello @YaSuenag, given these inputs, would you prefer updating this PR with the patch I showed in my previous comment or would you want me to open a separate PR with that patch? Either is fine with me.

YaSuenag:
@jonathan-gibbons Thanks for the info! I wonder why the issue on wget2 is still open, and it is not clarified whether wget2 guys will fix it or not.

Anyway, I agree with you to fix to remove usage of `..`. It is a big different from this PR. So @jaikiran could you open another PR with your patch? I will close this after you open.

jaikiran:
>  So @jaikiran could you open another PR with your patch? I will close this after you open.

I'll do it today. Thank you @YaSuenag.

jaikiran:
I've now opened https://github.com/openjdk/jtreg/pull/214

YaSuenag:
Thanks @jaikiran !

PR #216, opened 2024-07-26
7901757: Race in counting total number of failures from TestNG

jaikiran:
Can I please get a review of this change which proposes to fix the issue noted in https://bugs.openjdk.org/browse/CODETOOLS-7901757?

As noted there, the current implementation in jtreg's testng test listener doesn't take into account that the callback methods on that listener instance can be invoked concurrently for different threads. As a result, the count tracking logic in the listener ends up in a race condition which results in reporting incorrect numbers.

The commit in this PR fixes that issue and introduces a self test which reproduces the race issue and verifies the fix. The new test and existing tests continue to pass with this change.

---------
### Progress

### Issue
 * [CODETOOLS-7901757](https://bugs.openjdk.org/browse/CODETOOLS-7901757): Race in counting total number of failures from TestNG (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/216/head:pull/216` \
`$ git checkout pull/216`

`$ git checkout pull/216` \
`$ git pull https://git.openjdk.org/jtreg.git pull/216/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 216`

`$ git pr show -t 216`

</details>

<a href="https://git.openjdk.org/jtreg/pull/216.diff">https://git.openjdk.org/jtreg/pull/216.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/216#issuecomment-2252396201)

### DIFF
--- a/src/share/classes/com/sun/javatest/regtest/agent/TestNGRunner.java
+++ b/src/share/classes/com/sun/javatest/regtest/agent/TestNGRunner.java
@@ -34,6 +34,7 @@ import java.util.Collections;
 import java.util.IdentityHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
 import org.testng.IConfigurationListener;
@@ -104,8 +105,11 @@ public class TestNGRunner implements MainActionHelper.TestRunner {
         testng.addListener(new XMLReporter());
         testng.setOutputDirectory(new File(".").getPath()); // current dir, i.e. scratch dir
         testng.run();
-        if (listener.configFailureCount > 0 || listener.failureCount > 0) {
-            throw new Exception("failures: " + listener.failureCount);
+        int configFailureCount = listener.configFailureCount.get();
+        int testFailureCount = listener.failureCount.get();
+        if (configFailureCount > 0 || testFailureCount > 0) {
+            throw new Exception("config failures: " + configFailureCount
+                    + ", test failures: " + testFailureCount);
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/share/classes/com/sun/javatest/regtest/agent/TestNGRunner.java
```diff
@@ -104,22 +105,33 @@ public static void main(ClassLoader loader, String... args) throws Exception {
...
+                    + ", test failures: " + testFailureCount);
         }
     }
 
     public static class RegressionListener
             implements ITestListener, IConfigurationListener {
         enum InfoKind { CONFIG, TEST }
 
+        private final AtomicInteger count = new AtomicInteger();
+        private final AtomicInteger successCount  = new AtomicInteger();
```
pavelrappo, line 121:
```suggestion
        private final AtomicInteger successCount = new AtomicInteger();
```

pavelrappo (commented):
I'm not a Reviewer in this project, but I'm the submitter of the original bug. Thanks for finally addressing it. From my perspective, the approach looks correct and straightforward.

sormuras (approved):
Implementation and tests look good to me.

With soon correct numbers being reported, will we see new errors in the OpenJDK (and other projects') test suites?

jaikiran:
Hello Christian,

> With soon correct numbers being reported, will we see new errors in the OpenJDK (and other projects') test suites?

I don't expect the change in this PR to expose failures that have gone undetected so far, because the issue here is a race condition that could have incorrectly reported the failure count. For the failure to be not noticed previously, then it would mean that the race would have had to happen every single run of that test, which would be very odd.

In any case, in the next few days, I will include this change as part of the JDK tier testing I run as part of new jtreg changes and see how it goes.

jonathan-gibbons:
We should make sure that general jtreg documentation accurately describes when tests may or do not run concurrently

jonathan-gibbons (approved):

jaikiran:
> In any case, in the next few days, I will include this change as part of the JDK tier testing I run as part of new jtreg changes and see how it goes.

I tried this against the JDK mainline and although the run had failures those failures are unrelated to this change. I will go ahead and integrate this change later today.

PR #217, opened 2024-08-07
7903785: Improve reporting for skipped tests

jonathan-gibbons:
This is to support the ability to track `skipped` tests in the JDK `make test` system.

It is somewhere between "proof of concept", "draft PR", and initial version of a real PR.

The primary goal is to post the number of `skipped` tests in a backwards compatible way that does not break existing makefile rules or require any changes.  The makefiles currently have some "fuzzy parsing" to search for named values that may appear on the `Test results:` line near the end of the `jtreg` output to the console.   This PR _adds_ a new value to that line _without changing any existing values_. In particular, the `passed` value is unchanged and continues to include the `skipped` tests, as before, but for readers and tools alike, the number of skipped tests is also reported.

The PR also exposes a previously undocumented feature, to control the formatting of that line, using a format string that can be given to `jtreg` in a system property. The set of format specifiers in that string has been extended to include format specifiers for "skipped tests" and "passed tests, excluding skipped tests".

In exposing that feature, some design questions come to mind:
* the terminology of "ignored" tests is somewhat confusing -- it is about tests that were not selected by virtue of the keyword filter (and related `-k` option)
* if the number of `skipped` tests is interesting, there are other similar numbers that might also be interesting, such as the number of tests that were filtered out by a `@requires` tag -- which is similar to but more powerful than the keyword filter. But these extra numbers are contained within a composite filter, and may not have been so easily accessible when `jtreg` was built using older versions of JDK -- that is, before covariant returns.
* the `conditional space` and `conditional comma` provided as format specifiers do not work as well as intended if there is a plain-text label at the beginning of the format string.

But to summarize, this feature is primarily about reporting the number of `skipped` tests. We can either defer supporting additional values until a subsequent PR, or put more work into this PR to expose more of these "interesting" numbers.

---------
### Progress

### Issue
 * [CODETOOLS-7903785](https://bugs.openjdk.org/browse/CODETOOLS-7903785): Improve reporting for skipped tests (**Enhancement** - P3)

### Reviewers
 * [Daniel Jeliński](https://openjdk.org/census#djelinski) (@djelinski - no project role) ⚠️ Review applies to [affc138e](https://git.openjdk.org/jtreg/pull/217/files/affc138eaebbb697c3da563887f4de55e95c643d)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/217/head:pull/217` \
`$ git checkout pull/217`

`$ git checkout pull/217` \
`$ git pull https://git.openjdk.org/jtreg.git pull/217/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 217`

`$ git pr show -t 217`

</details>

<a href="https://git.openjdk.org/jtreg/pull/217.diff">https://git.openjdk.org/jtreg/pull/217.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/217#issuecomment-2274079552)


jaikiran:
Hello Jon, I haven't yet reviewed the code in detail, but have only looked at the PR description and the high level changes. With my limited knowledge of this project, the proposed idea looks good to me.

Does `excluded` cover only problem listed exclusions?

With the proposed change, I think we will be categorizing tests that haven't run the actual test logic into different categories - "excluded" (due to problem listing?), "skipped" (due to jtreg.SkippedException or test framework specific SkippedException type?), "ignored" (due to keyword evaluation). If later we do support "@requries" exclusion reporting, would that be a category of its own?

The other thing I think we should additionally consider is a summary report of these tests that haven't run. For example, if the count says 3 tests were skipped, 1 test was ignored, then I think we should have a notrun-summary.txt (keeping aside the file name for now) which would have something like:

```
skipped:
foo/bar/HelloTest.java threw jtreg.SkippedException
foo/bar/OneTwoThree.java threw testng.SkippedException
foo/bar/Five.java threw junit.AssumptionFailureException

ignored:
hello/world/HelloWorld.java  due to keyword "intermittent"
```

jonathan-gibbons:
I just encountered an "unexpected" skipped test. This is good.

It made me think that we should modify the "verbose" output (especially -v1) to indicate if a test has been skipped.

jonathan-gibbons:
@jaikiran 
I like the idea of a simple `not-run-summary.txt` file.

These files tend to be line-oriented, for easy manipulation with `grep`, so I think the output should be a series of lines of the form: `reason: file`

The reason would be simple, just indicating the "filter" that caused the test to be not run. It would not be practical to dig into the specific reasons that a keyword expression or requires expression cause the test to be not run.

jaikiran:
The format that you note as the content in `not-run-summary.txt` looks good to me. I think that should be enough to do any further investigation to understand why a test wasn't run for that given reason.

sormuras:
Perhaps we should come up with a better name capturing the gist of:
> While we provide a low-level mechanism for skipping a test, by converting a stylized exception to a stylized `passed` result, we do not reflect that at the top level reporting.

While adding the feature to write the `not-run-summary.txt` report I was eluded to check for `TestResult.getStatus().isNotRun()` as the name matches so nicely - but we're not really interested in those results returning `true` here.

What about `skipped-reasons.txt` ?

jonathan-gibbons:
I've been thinking about this some more, and I'm not sure how much we need 
the work as-is, and/or what the overall desired functionality should be.

Tests may be selected for execution in a number of ways; equivalently,
they may be filtered out from selection in the same number of ways.
Here are the ways that tests can be selected or filtered out.

* The initial set of tests to be executed is given by the set of 
  initial files or groups, given as paths on the command-line.

* The tests are then passed through a series of filters to
  select or reject tests to be executed. Most of these are
  relatively simple filters, specified by command-line options.
  The order in which these filters are applied is not specified.
  A test must be accepted by all these filters to be executed;
  otherwise, it is "first one wins" to reject a test if it should
  not be executed.

  * _exclude list_ -- exclude tests given in "problem-list" files,
    given with `-exclude` on the command line
  * _keywords_ -- only run tests matching a keyword expression 
    given with `-k` on the command line
  * _prior status_ -- only run tests whose prior execution status 
    matches those given on the command line with `-status`
  * _time limit_ -- only run those tests that are expected to take 
    less time than a value given on the command line with `-tl`
    or `-timelimit`
  * _match list_ -- only tests given in "problem-list" files,
    given with `-match` on the command line; used to explicitly
    run tests that are normally excluded

The last filter in this group is always enabled and is not
directly controlled by command-line options.

* _requires_ -- check whether the test statically meets the
  conditions met in a `@requires` tag in the test description.
  This may use characteristics (properties) of the test VM, 
  such as the OS name, and family, the max memory size and 
  so on. Some of the properties may be exported from the
  underlying VM.  The expression is evaluated by `jtreg`;
  the test itself is not executed at this stage.

If a test given by one of the initial files or test groups
is accepted by all the preceding filters, it will be executed.
When it is executed, there are various possible outcomes.

* The test may dynamically determine that it cannot be executed
  as intended. If this occurs, it should indicate that outcome
  by throwing `jtreg.SkippedException`, which will be treated 
  by the harness as a "skip": not passed but not failed either".
  This can be thought of as a "dynamic" filter, similar to the
  static "requires" filter. The difference is that the check
  happens within the running test.

* A combo or framework test may determine that some parts of the 
  cannot be excuted as intended. If this occurs, the framework
  should report that in some appropriate manner, and then
  report passed or failed for the parts that it can execute.

* The test should report passed or failed as appropriate.

So, where does that leave this work.

1. Because skipped tests (those that throw `jtreg.SkippedException`)
   are actually executed, the information about those tests
   are recorded in the exeisting `report/text/summary.txt` file.
   Look for the lines containing `Passed. Skipped: jtreg.SkippedException:`.
   Count the tests with `wc -l`; report them with `grep`.
   Thus, it is of lesser importance to report such information
   by itself.

2. There are other numbers related to other reasons why a test might
   not be executed that may also be interesting. Of these, maybe
   the information from the 'requires' filter is the most interesting
   because it is the least obvious. and the least explicitly dependent 
   on command-line options.

jonathan-gibbons:
Big new update.

In terms the original request, to report "skipped" tests when running tests with `make test`, there is almost no need to do anything, since the information is already available, if you know where to look. That being said, it is still a convenience to provide the information, to benefit anyone running `jtreg` directly, instead of via the makefiles.

But, in reviewing the conversations that led to this work, the underlying interest was more in terms of knowing and understanding the lack of test coverage in an area, which includes the tests that were not run at all, as well as just those that were started and then skipped.

So, this update provides detailed information on all the tests that were not run, as well as the tests that were run (and possibly skipped). The tests that are not run are those that were specified in the "initial files" (including groups), but which were subsequently filtered out by one of the various filters.

I gave up trying to find meaningful single words that describe the reasons that a test might be filtered out, and went with short phrases instead.  In addition, the `jtreg` FAQ is updated which more detailed explanations, to help anyone who is search for any such details.

In terms of formatting the numbers, I chose to keep all the labeled numbers on a single line, just because there is a spectrum of conditions here that got in the way of splitting the info into many lines: _tests that were run_, _tests that were run but skipped_, _tests that were not run_.

*In addition,*
A new file `text/notRun.txt` is generated in the report directory that uses the same format as `text/summary.txt` but which lists the primary reason why each test was not executed -- where _primary_ means _first_ in the unspecified order in which the filters are applied.  For the record, I tried quite hard to include the information in `summary.txt` but although I succeeded at one point, the results were somewhat unsatisfactory, not least because it meant overwriting and results for a test in the work directory.

*Notes,*
1. I (re)discovered the `jtreg` `-report` option, which can be used to control the set of tests in the report that is generated at the end of a test run. In conjunction with the effectively undocumented support for `JT Harness` preferences, It is possible to have the system report all the filters that rejected any individual test, not just the first one. But the filters are identified by names that are neither consistent nor "user-friendly".
2. In documenting the use of the keywords filter, I realize that we do not seem to adequately document the "implicit" keywords defined by any test description, not the "effective" keyword expression created from the `jtreg` command-line options. But that is another RFE, for another day.

test/basic/ReportOnlyTest.gmk
```diff
@@ -29,7 +29,7 @@ ifdef HEADLESS
 REPORT_EXPECT_PASS = 91
```
sormuras, line 29:
```suggestion
REPORT_EXPECT_PASS = 92
```

At least on GHA there seem to be 92 tests that pass.

jonathan-gibbons: I'll have to check your suggested change; my recollection is that all tests passed when run locally.

jonathan-gibbons: The warnings are as expected when conditions are as found.

sormuras: > I'll have to check your suggested change; my recollection is that all tests passed when run locally.

The `else` branch (`HEADLESS != 1`) also has an incremented `REPORT_EXPECT_PASS` value: from `93`  to `94`.

Note that the merge with `upstream/master` still leads to the same error:
```
/bin/grep -s "Test results: passed: ${EXPECT_PASS}; failed: ${EXPECT_FAIL}; error: 88" /home/runner/work/jtreg/jtreg/build/test/ReportOnlyTest.jt.log  > /dev/null
Expect: Test results: passed: 91; failed: 40; error: 88
make: *** [../test/basic/ReportOnlyTest.gmk:45: /home/runner/work/jtreg/jtreg/build/test/ReportOnlyTest.ok] Error 1
```

LudwikJaniuk:
I've been pulled into other work, but am very glad to see all the movement here. Very happy to see the direction change from just focusing on dynamically skipped tests to a wider-boundary consideration. I intend to review this within the next week.

src/share/doc/javatest/regtest/faq.md
```diff
@@ -900,6 +900,123 @@ You have several alternatives.
...
+:   The number of tests that were executed and which indicated that the test passed.
+    Note that some parts of the test may not have been executed. Some tests may be "skipped",
+    meaning that the test determined that it could not be executed as intended.
+
+_failed_
+:   The number of tests that were executed and which indicated that the test failed.
+
+_error_
+:   The number of tests that were executed and which indicated that an error
+    occurred bedfore it could be determined whether the test passed or failed.
```
Domest0s, line 961:
```suggestion
    occurred before it could be determined whether the test passed or failed.
```
Probably a typo.

Domest0s (commented):
Would it be possible to generate a report for every test which was skipped? I am not sure if that's already possible or will be possible with this change. It's likely out of scope.
But it would be nice to know your thoughts.

jonathan-gibbons:
> Would it be possible to generate a report for every test which was skipped? I am not sure if that's already possible or will be possible with this change. It's likely out of scope. But it would be nice to know your thoughts.

I don't know what that means, or what you are asking for.

Partly to clarify, in this context, "skipped" means a test was started, but the test dynamically determined it could not proceed, and so threw a `jtreg.SkippedException`.  Because the test was started, there will be a `.jtr` file giving the standard details of how the test was compiled and executed, and will contain any output written by the test up to the point it threw the exception.

Separately, some tests are "not run" (in this context, that's not the same as "skipped").  A test may be not run for many reasons, but all generally of the form that the test was not accepted by one of the various filters, such as the keyword filter, the exclude-list filter, the `requires` filter, and so on. If a test is "not run" there is nothing additional that can reasonably be reported. Yes, in principle, `jtreg` could report fine-grain detail of how and why whichever filter(s) rejected the test, but architecturally, that is not really possible.

To summarize, the `notRun.txt` file will contain a brief one-line summary of why a test was not run. So, you will be able to see whether it was an issue with keywords, requires, modules, problem list, and so on.

djelinski:
@jaikiran example notRun.txt file content:
```
applet/Fail.java#id1   Not run. Not matching the given keyword expression: !manual
applet/Fail.java#id2   Not run. Not matching the given keyword expression: !manual
```
still running tests using the new jtreg, looks good so far. Will report when done.

tstuefe:
I don't know the code and can therefore not review, but I welcome this work. This is very useful and I have wanted this feature for a long time.

src/share/doc/javatest/regtest/faq.md
```diff
@@ -900,6 +900,123 @@ You have several alternatives.
...
+* `%?X`        &mdash; prints given number if not zero, where X is one of f, F, e, p, P, s, x, i
+* `%?{textX}`  &mdash; prints text and given number if number is not zero, where X is one of f, F, e, p, s, x, i
+
+A _conditional_ comma or space is only generated if it is not at the beginning of the line.
+
+### What do all those numbers in the "Test results" line mean?
+
+After running tests, `jtreg` prints out a line beginning `Test sesults:` followed
+by a series of labeled numbers. The numbers give details about the number of
+tests that were as well as the number of tests that were not. What do they all mean?
```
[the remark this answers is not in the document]
jonathan-gibbons: Agreed. Will fix.

djelinski (approved):

sormuras:
Some basic counters don't match: https://github.com/openjdk/jtreg/pull/217#discussion_r1734439374 -- perhaps related to GHA be running with Java 17?

jonathan-gibbons:
> Some basic counters don't match: [#217 (comment)](https://github.com/openjdk/jtreg/pull/217#discussion_r1734439374) -- perhaps related to GHA be running with Java 17?

or related to headless mode, again

jonathan-gibbons:
using `make -k test` to run all tests, and find which do not pass, all tests except `LibPropertiesEnablePreviewTest.ok` pass, forme: macOS, JDK 11, etc

jonathan-gibbons:
At this time, 2:48 PDT, all GHA tests based. Yay!

sormuras (approved):

jaikiran (approved):

test/basic/ReportOnlyTest.gmk
sormuras: Fixed via https://github.com/openjdk/jtreg/pull/217/commits/5c990293d97e15d4b292052210f977338640df74

PR #225, opened 2024-09-02
7903813: Tag enablePreview with false should take precedence

sormuras:
Please review this change to let tests with `@enablePreview false` take precedence over libraries that require preview features being enabled.

Note that the [PreviewHiddenClass](https://github.com/openjdk/jdk/blob/f6d7e30b84fedbf42077526610ba7a5bcfaece4c/test/jdk/java/lang/invoke/defineHiddenClass/PreviewHiddenClass.java#L34-L35) test should receive an explicit `enablePreview=false` tag line (in addition to the comment).

---------
### Progress

### Issue
 * [CODETOOLS-7903813](https://bugs.openjdk.org/browse/CODETOOLS-7903813): Tag enablePreview with false should take precedence (**Bug** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/225/head:pull/225` \
`$ git checkout pull/225`

`$ git checkout pull/225` \
`$ git pull https://git.openjdk.org/jtreg.git pull/225/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 225`

`$ git pr show -t 225`

</details>

<a href="https://git.openjdk.org/jtreg/pull/225.diff">https://git.openjdk.org/jtreg/pull/225.diff</a>

</details>

### Webrev
[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/225#issuecomment-2324893628)

### DIFF
--- a/src/share/classes/com/sun/javatest/regtest/exec/BuildAction.java
+++ b/src/share/classes/com/sun/javatest/regtest/exec/BuildAction.java
@@ -312,7 +312,8 @@ public class BuildAction extends Action
                     // "--enable-preview" requires either "--source" or "--release" to
                     // confirm the expected source level. "--release" restricts
                     // the visible API to the exported API, so use "--source" instead.
-                    compArgs.add("--source=" + script.getTestJDKVersion().major);
+                    compArgs.add("-source"); // use "-source" "N" for Java 11 and lower
+                    compArgs.add(String.valueOf(script.getTestJDKVersion().major));
                 }
             } catch (UncheckedIOException exception) {
                 throw new TestRunException("Reading library properties failed: " + libLocn, exception);
--- a/src/share/classes/com/sun/javatest/regtest/exec/CompileAction.java
+++ b/src/share/classes/com/sun/javatest/regtest/exec/CompileAction.java
@@ -366,7 +366,7 @@ public class CompileAction extends Action {
         if (runJavac
                 && !script.disablePreview()
                 && !seenEnablePreview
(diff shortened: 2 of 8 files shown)


jonathan-gibbons:
Locally, some tests fail.

jonathan-gibbons:
You need the following additional two-line patch:
[patch.txt](https://github.com/user-attachments/files/16937030/patch.txt)


If you can't see that, look in `BuildAction.java`, line 315, change to generate two values: `-source`, `N`, instead of `--source=N`.  Note the single dash form for `-source`.  The `--source` form is a more recent form, and does not work for JDK 9, 10, 11.

jonathan-gibbons (approved):
OK. I probably would not have written the code that way, and might have done more to  emphasize the three-valued nature (unset, `true`, `false`) of `enablePreview` in the `TestDescription`.  I did have to go (re)read the code for `RegressionTestFinder` to see how/where that an empty string for `@enablePreview` was handled.  But part of that is on me, I guess, for the original code in this area.

sormuras:
Thank you for the review, Jon.

> Maybe you can change the GHA tests to use JDK 11 for tests.

I'll look into that - ideally, we would have all those JDKs installed/available on GHA when running `jtreg`'s self-tests:

- Warning: JDK8HOME not set; some tests may not have been executed
- Warning: JDK9HOME not set; some tests may not have been executed
- Warning: JDK10HOME not set; some tests may not have been executed
- Warning: JDK14HOME not set; some tests may not have been executed
- Warning: JDK18HOME not set; some tests may not have been executed

PR #235, opened 2024-11-06
7903519 : jtreg/jtharness is missing features for basic crash testing

andrlos:
provides SPI for enabling external status transformations of failed tests

this is a continuation of efforts after https://github.com/openjdk/jtharness/pull/59 

Requires newest jtharness build (not even tagged yet) that includes above mentioned change to be compiled succesfully

The main idea is to provide a unified StatusTransformer interface, that can be externally implemented by users and added to a classpath in a separate jar to allow modifications of test execution status based on some elementary analysis. This can be easily used for crashtesting (filtering out only tests with jvm crashes).

---------
### Progress

### Issue
 * [CODETOOLS-7903519](https://bugs.openjdk.org/browse/CODETOOLS-7903519): jtreg/jtharness is missing features for basic crash testing (**Enhancement** - P3)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/235/head:pull/235` \
`$ git checkout pull/235`

`$ git checkout pull/235` \
`$ git pull https://git.openjdk.org/jtreg.git pull/235/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 235`

`$ git pr show -t 235`

</details>

<a href="https://git.openjdk.org/jtreg/pull/235.diff">https://git.openjdk.org/jtreg/pull/235.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/235#issuecomment-2458539420)
</details>

### DIFF
--- a/src/share/classes/com/sun/javatest/regtest/exec/RegressionScript.java
+++ b/src/share/classes/com/sun/javatest/regtest/exec/RegressionScript.java
@@ -51,6 +51,8 @@ import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 import java.util.stream.Collectors;
+import java.util.Iterator;
+import java.util.ServiceLoader;
 
 import com.sun.javatest.Script;
 import com.sun.javatest.Status;
@@ -79,6 +81,8 @@ import com.sun.javatest.regtest.report.Verbose;
 import com.sun.javatest.regtest.tool.Version;
 import com.sun.javatest.regtest.util.FileUtils;
 import com.sun.javatest.regtest.util.StringUtils;
+import com.sun.javatest.regtest.exec.StatusTransformer;
+
 
 import static com.sun.javatest.regtest.RStatus.error;
 import static com.sun.javatest.regtest.RStatus.passed;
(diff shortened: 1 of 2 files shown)


andrlos:
@sormuras 	feast your eyes on the code below :D

```
public class CrashOnlyStatusTransformer implements StatusTransformer {
        @Override
        public Status transform(Status originalStatus, TestDescription td) {
            Status newStatus = originalStatus;
            if(originalStatus.getType() == Status.FAILED && ! this.didCrash(td)){
                newStatus = new Status(Status.PASSED, "Just a regular failure.");
            }
            return newStatus;
        }

        private boolean didCrash(TestDescription td){
            Pattern pattern = Pattern.compile("^hs_err_pid(\\d+)\\.log");
            List<String> hs_errs = Arrays.stream(td.getDir().list()).filter(pattern.asPredicate()).collect(Collectors.toList());
            return !hs_errs.isEmpty();
    }
}
```

this is an approach that we use for crashtesting with debug jdk builds to separate crashes from regular failures

jaikiran:
Hello @andrlos, looking at that code you pasted:
```java
if(originalStatus.getType() == Status.FAILED && ! this.didCrash(td)){
    newStatus = new Status(Status.PASSED, "Just a regular failure.");
}
```
It looks odd to be marking a failed test as successful. Furthermore, doesn't a crashed JVM result in test status to be `Status.ERROR`?

andrlos:
@sormuras it proved to be much harder to filter them after, as the hs_err_pid log is not being copied, plus we have many tests that run via script, so a generic jvm watcher is not an option as would be the option with jvm agent that we use for junit testing.. we need to cover cases, where jvm crashes even tho it has been run via a shell script (or multiple levels of scripts) and watching for hs_err_pid.log proved to be the most reliable and universal approach.

andrlos:
I believe that this approach is pretty minimalistic when all the capabilities for the user are taken into account and presents almost no maintenance overhead. Plus the user of the jtreg project is someone who should know what he is doing and giving him tools to peek inside and modify the test evaluation behavior is nothing bad IMO. Default behavior of jtreg release is also not affected by this change.

andrlos:
@jaikiran @sormuras WDYT?

sormuras:
With jtreg 7.5.1 being released the other day and soon to be made default in JDK mainline, I will pick up the thread here soon again. Sorry for the delay.

jaikiran:
Hello @andrlos, sorry to keep you waiting. I don't have a strong opinion or objection to this change. At the same time, I don't have enough knowledge of this code to approve this change. So other than asking some basic questions (which I did), I don't feel qualified enough to officially review this proposed enhancement.

sormuras:
The use-case doesn't carry the weight of introducing and maintaining a SPI for arbitrary and distant test result mangling in the inards of `jtreg`. The aforementioned JUnit `TestRule` is a good example for a non-distant (think: test-local) way to influnce the result of a test in an understandable way.

In the light of the discussion above, I won't approve this pull request.

Let's discuss other approaches to achieve the goals expressed in https://bugs.openjdk.org/browse/CODETOOLS-7903519

andrlos:
Hi @sormuras do you have any ideas about possible approach? It seems that I either lost my openjdk login or I never had one to start with.. I will try and get it while you can start the discussion there. Thanks!

judovana:
@andrlos  please refresh/negotiate your openjdk jira system login. Lets continue the discussion here as sormuras required. Thaxn all!

sormuras:
As this thread already contains most of the related discussion, related topics, and affected people, let's continue here. Traces are recorded automatically to https://mail.openjdk.org/mailman/listinfo/jtreg-dev 's archives.

andrlos:
Okay, so the main idea would be to integrate a mechanism similar to JUNIT's  TestRule.. Do you have any suggestions of where you would see such a feature? I already proposed a way to integrate something similar into the RegressionScript code and further propagate it into the underlying JTHarness.. I am afraid that we can't really avoid propagation into jtharness to get that local non-distant approach.

sormuras:
What I still don't get is that you want to address a global goal "basic crash testing" with a very local mechanism, like something similar to JUnit 4 "TestRule".

I think a solution lies in between your lines:

> it proved to be much harder to filter them after, as the `hs_err_pid` log is not being copied, plus we have many tests that run via a shell script, so a generic jvm watcher is not an option as would be the option with jvm agent that we use for junit testing.. we need to cover cases, where jvm crashes even tho it has been run via a shell script (or multiple levels of scripts) and watching for `hs_err_pid.log` proved to be the most reliable and universal approach.

So, what about making sure that `jtreg` copies `hs_err_pid` files correctly?

andrlos:
@sormuras 	can definitely be tried, I did want to do something with minimal code changes as I saw it as a best chance to get it approved. Plus the TestRule approach is a very powerful way of doing so much more and could be used for other purposed in the future as well. However I can definitely give hs_err_pid files copying a look and see. Hopefully this wont break anything else.
Thanks for this short brainstorming, will get back to you, once I get a PoC working.

PR #237, opened 2024-12-14
7903920: Support building jtreg on FreeBSD

snake66:
On FreeBSD we need to use GNU gmake instead of the BSD make program.

---------
### Progress

### Issue
 * [CODETOOLS-7903920](https://bugs.openjdk.org/browse/CODETOOLS-7903920): Support building jtreg on FreeBSD (**Enhancement** - P4)

### Reviewers
 * [Magnus Ihse Bursie](https://openjdk.org/census#ihse) (@magicus - Author)
 * [Jan Lahoda](https://openjdk.org/census#jlahoda) (@lahodaj - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/237/head:pull/237` \
`$ git checkout pull/237`

`$ git checkout pull/237` \
`$ git pull https://git.openjdk.org/jtreg.git pull/237/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 237`

`$ git pr show -t 237`

</details>

<a href="https://git.openjdk.org/jtreg/pull/237.diff">https://git.openjdk.org/jtreg/pull/237.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/237#issuecomment-2613356188)
</details>

### DIFF
--- a/make/Defs.gmk
+++ b/make/Defs.gmk
@@ -182,9 +182,17 @@ else
 TIDY	= /usr/bin/tidy
 endif
 TOUCH 	= /usr/bin/touch
+ifeq ($(SYSTEM_UNAME), FreeBSD)
+UNZIP	= /usr/local/bin/unzip
+else
 UNZIP 	= /usr/bin/unzip
+endif
 WC 	= /usr/bin/wc
+ifeq ($(SYSTEM_UNAME), FreeBSD)
+ZIP	= /usr/local/bin/zip
+else
 ZIP 	= /usr/bin/zip
+endif
 
 
 #----------------------------------------------------------------------
--- a/make/Defs.gmk
+++ b/make/Defs.gmk
@@ -141,7 +141,15 @@ CP 	= /bin/cp
 DIFF 	= /usr/bin/diff
 ECHO	= /bin/echo
 FIND	= /usr/bin/find
+ifeq ($(SYSTEM_UNAME), FreeBSD)
+GREP  := $(shell if [ -r /usr/local/bin/ggrep ]; then \
+    echo /usr/local/bin/ggrep ; \
+  else \
+    echo $(ECHO) "GNU grep is required, but was not found" ; \
+  fi )
+else
 GREP 	:= $(shell if [ -r /bin/grep ]; then echo /bin/grep ; else echo /usr/bin/grep ; fi )
+endif
 LN	= /bin/ln
 LS	= /bin/ls
 MKDIR 	= /bin/mkdir
--- a/make/Platform.gmk
+++ b/make/Platform.gmk
@@ -56,6 +56,11 @@ ifeq ($(UNAME_S), Linux)
   endif
   OS_VERSION := $(shell $(UNAME) -r)
 endif
+ifeq ($(UNAME_S), FreeBSD)
+  OS_NAME     = freebsd
+  OS_ARCH    := $(shell $(UNAME) -m)
+  OS_VERSION := $(shell $(UNAME) -r)
+endif
 ifeq ($(UNAME_S), Darwin)
   OS_NAME     = macosx
   OS_ARCH    := $(shell $(UNAME) -m)
--- a/make/build.sh
+++ b/make/build.sh
@@ -765,13 +765,23 @@ if [ -n "${SKIP_MAKE:-}" ]; then
     exit
 fi
 
+MAKE=$(which make)
+setup_make() {
+    case `uname` in
+      FreeBSD )
+          MAKE=/usr/local/bin/gmake ;;
+    esac
+}
+setup_make
+check_file "${MAKE}"
+
 # save make command for possible later reuse, bypassing this script
 mkdir -p ${BUILD_DIR}
 cat > ${BUILD_DIR}/make.sh << EOF
 #!/bin/sh
 
 cd "${ROOT}/make"
-make ASMTOOLS_JAR="${ASMTOOLS_JAR}"                           \\
+${MAKE} ASMTOOLS_JAR="${ASMTOOLS_JAR}"                           \\
      ASMTOOLS_NOTICES="${ASMTOOLS_NOTICES}"                   \\
      BUILDDIR="${BUILD_DIR}"                                  \\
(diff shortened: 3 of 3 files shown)


snake66:
This work is done on behalf of The FreeBSD Foundation. https://oca.opensource.oracle.com/api/v1/members/public?&search=FreeBSD

sormuras:
Thanks for creating the PR, @snake66 - can you please create a JBS issue for this change? Happy to help if needed: https://bugs.openjdk.org/browse/CODETOOLS

@jonathan-gibbons wrote in https://github.com/openjdk/jtreg/commit/8a66b3b8f8f844f442a74f1d651cfc03f9231bb7#commitcomment-150335752
> Stylistically, it would be more in keeping to determine the absolute path on FreeBSD, instead of just relying on $PATH. You can do this either by checking the OS name or by checking in standard well-known places in order to determine an absolute path. See examples of other variables nearby.

Seems like this change avoids relying on `$PATH`. Do you also consider doing something similar for `zip` and `unzip` in this change set?

snake66:
> Thanks for creating the PR, @snake66 - can you please create a JBS issue for this change? Happy to help if needed: https://bugs.openjdk.org/browse/CODETOOLS

Thanks, I'm not sure if I have access to create issues? The credentials for the oracle account did not work.

> Seems like this change avoids relying on `$PATH`. Do you also consider doing something similar for `zip` and `unzip` in this change set?

Yes, I was unsure if those changes would pass, so I'll follow your recommendations.

Do you want me to include the rest of the FreeBSD related changes to this PR, or is it preferable to do separate PR's for each change?

sormuras:
Created https://bugs.openjdk.org/browse/CODETOOLS-7903920 to track this improvement to `jtreg`'s build system "in one go".

make/Defs.gmk
```diff
@@ -141,7 +141,15 @@ CP 	= /bin/cp
 DIFF 	= /usr/bin/diff
 ECHO	= /bin/echo
 FIND	= /usr/bin/find
+ifeq ($(SYSTEM_UNAME), FreeBSD)
+GREP  := $(shell if [ -r /usr/local/bin/ggrep ]; then \
```
magicus, line 145: Maybe consider using ``which ggrep`` instead?

make/Defs.gmk
```diff
@@ -174,9 +182,17 @@ else
 TIDY	= /usr/bin/tidy
 endif
 TOUCH 	= /usr/bin/touch
+ifeq ($(SYSTEM_UNAME), FreeBSD)
+UNZIP	= /usr/local/bin/unzip
```
magicus, line 186: Once again, maybe use `which` instead? Unless you have two installations of unzip, and the one in /usr/bin is broken in some way. Then you might need to check for brokenness instead.

make/Defs.gmk
snake66: Wouldn't this pretty much just be the same as just not specifying the path at all, and let the shell find it in $PATH? Which was [one of the complaints](https://github.com/openjdk/jtreg/pull/237#issuecomment-2543058251) against my first patch.

magicus: Okay, people have apparently varying opinions on this. :) Don't want to put you in a "damned if you do, damned if you don't" situation, so just ignore my comment and go with your solution.

snake66: I don't have any strong opinions about this, but for FreeBSD's part it would actually be slightly better to let the shell find the binaries in $PATH. But it's not important for my progress now, so I figured I'd deal with it if it comes up :)

jonathan-gibbons: Maybe the concern is becoming obsolete, but the general design focus of this and similar scripts was to lock down everything, and avoid any variation that might occur as a result of individuals setting non-standard values for PATH and similar path-related variables -- CLASSPATH anyone?

But, a reasonable compromise would be to limit the use of `which` to this file (`Defs.gmk`) and maybe to within suitable `ifeq ($(SYSTEM_UNAME), FreeBSD)` constructs.

That all being said, I would defer to the expertise and style-advice of folk like @magicus on the Build team,  but with some amount of caveat for overall style consistency, as much as is reasonable.

magicus: If the path is hardcoded as e.g. `/usr/bin/foo`, I tend to agree with you. That means that an official system tool is needed and confirmed working. However, hard-coding `/usr/local/bin/foo` is another matter altogether. It means there is no standard system tool that is working, and we presume the user has installed a "correct" tool locally.

magicus: If the path is hardcoded as e.g. `/usr/bin/foo`, I tend to agree with you. That means that an official system tool is needed and confirmed working. However, hard-coding `/usr/local/bin/foo` is another matter altogether. It means there is no standard system tool that is working, and we presume the user has installed a "correct" tool locally.

magicus: (uness BSD does thing different in this regard and install all user-installed packages in /usr/local?)

snake66: > (uness BSD does thing different in this regard and install all user-installed packages in /usr/local?)

Yep, that's how FreeBSD does things. (Technically the prefix can be something different than /usr/local, but that's the default.)

magicus: Ok, in that case this makes totally sense.

magicus (approved):
Looks good to me. (But I'm no Reviewer)

sormuras:
@jonathan-gibbons Do you have any additional comments or changes to suggest/request? If not, a final **R**eview of this PR would be awesome. 🤓

lahodaj (approved):
Although I am not really an expert, but since @magicus approved, looks sensible to me.

snake66:
Thank you @lahodaj

sormuras (approved):

sormuras:
> ➡️ To flag this PR as ready for integration with the above commit message, type /integrate in a new comment. (Afterwards, your sponsor types /sponsor in a new comment to perform the integration).

Happy to sponsor your PR, after you commented with `/integrate`, @snake66.

PR #241, opened 2025-01-13
7903930: Support running individual parameterized tests and @Nested test classes

JornVernee:
See the JBS issue for an extended problem description.

This PR adds support for running individual parameterized JUnit test methods, as well as nested test classes, by extending the format that a query string can have. The new proposed format is as follows:

```
<class name>[::<method name>([<param type>[...,<param type>]])]
```

For example, running a method `foo` that takes no parameters would be done using `<test name>?TestClass::foo()`. If `foo` has parameters, then at least for JUnit tests they have to be specified as a comma separated list of binary names between the parentheses. For example:

```
TestDowncallScope::testDowncall(int,java.lang.String,CallGeneratorHelper$Ret,java.util.List,java.util.List)
```

The delimiting characters I choose here `::` and `(...)` have to be quoted when passed through most shells. If this is deemed to be an issue, we could also change the separators for class name/method name/parameters to simple hyphens. Although, in cases where a `$` sign appears in the binary name of a nested class, quoting would be needed either way. (I would personally choose to always quote the argument, just to be sure, so I don't mind using `::` and `(...)`).

Testing, I've added additional tests for the new cases (as far as I know, TestNG does not support nested test classes though). I've also been using a PoC of this feature as the basis for similar support in the intellij plugin that I'm working on.

### DIFF
--- a/src/share/classes/com/sun/javatest/regtest/agent/JUnitRunner.java
+++ b/src/share/classes/com/sun/javatest/regtest/agent/JUnitRunner.java
@@ -46,11 +46,13 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.lang.reflect.Method;
 import java.time.Duration;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.stream.Collectors;
 
 /**
  * TestRunner to run JUnit tests.
@@ -127,13 +129,27 @@ public class JUnitRunner implements MainActionHelper.TestRunner {
         // https://junit.org/junit5/docs/current/user-guide/#launcher-api-execution
         Thread.currentThread().setContextClassLoader(mainClass.getClassLoader());
         try {
-            // if test.query is set, treat it as a method name to be executed
-            String testQuery = System.getProperty("test.query");
-            LauncherDiscoveryRequest request = LauncherDiscoveryRequestBuilder.request()
-                .selectors(testQuery == null
-                        ? DiscoverySelectors.selectClass(mainClass)
-                        : DiscoverySelectors.selectMethod(mainClass, testQuery))
-                .build();
+            String testQueryStr = System.getProperty("test.query");
+            LauncherDiscoveryRequest request;
+            if (testQueryStr != null && !testQueryStr.isEmpty()) {
+                TestQuery testQuery = TestQuery.parse(testQueryStr);
+                String className = testQuery.className();
+                Optional<String> methodName = testQuery.methodName();
+                Optional<List<String>> paramTypeNames = testQuery.paramTypeNames();
+                Optional<String> paramTypeNamesStr = paramTypeNames.map(l -> l.stream().collect(Collectors.joining(",")));
+
+                request = LauncherDiscoveryRequestBuilder.request()
+                    .selectors(methodName.isPresent()
+                            ? paramTypeNamesStr.isPresent()
+                                ? DiscoverySelectors.selectMethod(className, methodName.get(), paramTypeNamesStr.get())
+                                : DiscoverySelectors.selectMethod(className, methodName.get())
+                            : DiscoverySelectors.selectClass(className))
+                        .build();
+            } else {
+                request = LauncherDiscoveryRequestBuilder.request()
+                    .selectors(DiscoverySelectors.selectClass(mainClass))
+                    .build();
+            }
 
             SummaryGeneratingListener summaryGeneratingListener = new SummaryGeneratingListener();
 
--- a/make/Defs.gmk
+++ b/make/Defs.gmk
@@ -173,6 +173,7 @@ TIDY 	:= $(shell if [ -r /usr/local/bin/tidy ]; then \
 else
 TIDY	= /usr/bin/tidy
 endif
+TR      = /usr/bin/tr
 TOUCH 	= /usr/bin/touch
 UNZIP 	= /usr/bin/unzip
 WC 	= /usr/bin/wc
--- a/src/share/classes/com/sun/javatest/regtest/agent/TestNGRunner.java
+++ b/src/share/classes/com/sun/javatest/regtest/agent/TestNGRunner.java
@@ -30,10 +30,7 @@ import java.io.PrintWriter;
 import java.io.StringWriter;
 import java.lang.reflect.Method;
 import java.time.Duration;
-import java.util.Collections;
-import java.util.IdentityHashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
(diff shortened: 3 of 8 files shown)


sormuras:
Implementation and associated tests look good on first sight! Will do some local experiments tomorrow.

Please, also update related sections/phrases in the FAQ document, for example: 
- https://github.com/openjdk/jtreg/blob/master/src/share/doc/javatest/regtest/faq.md#how-do-i-specify-which-tests-to-run
- https://github.com/openjdk/jtreg/blob/master/src/share/doc/javatest/regtest/faq.md#how-do-i-run-a-single-test-method-in-a-junit-test

JornVernee:
I've updated the FAQ. See: https://github.com/openjdk/jtreg/pull/241/commits/54d91bde8fdc0993866dc3cd1886909dec38ab7e

JornVernee:
Some results of offline discussion:

- We'd like to rely on JUnit's builtin support for parsing these queries, so we don't have to implement our own parsing. This also means we can use any JUnit selector (such as uid, to run individual cases of a parameterized test)
- This means that we don't support the extended format for TestNG tests. But, this is okay, since TestNG doesn't support nested test classes any way, and we don't need to specify the parameter types to select a parameterized method
- To use the junit selector support, the query string is prefixed with `junit-select:`, and then an identifier such as the ones listed [here](https://junit.org/junit5/docs/current/user-guide/#running-tests-discovery-selectors) is used as a suffix.
- The current format, with just the method name, continues to work as well.

src/share/classes/com/sun/javatest/regtest/agent/JUnitRunner.java
```diff
@@ -46,11 +47,13 @@
 import java.io.StringWriter;
 import java.lang.reflect.Method;
 import java.time.Duration;
+import java.util.List;
```
sormuras, line 50: Remove unused import statement.

```suggestion
```

src/share/classes/com/sun/javatest/regtest/agent/JUnitRunner.java
```diff
@@ -46,11 +47,13 @@
...
 import java.util.Map;
 import java.util.Optional;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.locks.Lock;
 import java.util.concurrent.locks.ReentrantLock;
+import java.util.stream.Collectors;
```
sormuras, line 56: Remove unused import statement.

```suggestion
```

src/share/classes/com/sun/javatest/regtest/config/TestManager.java
```diff
@@ -101,10 +101,15 @@ public static class GroupSpec {
...
+         * Additionally, since query strings are indicated by '?', and can contain
+         * any string, the '?' character is not allowed to be in the file path, to
+         * avoid picking up tests with a query string that contains ':' as being
+         * group specs.
          */
         static final Pattern groupPtn = System.getProperty("os.name").matches("(?i)windows.*")
-                ? Pattern.compile("(?<dir>|[^A-Za-z]|.{2,}):(?<group>[A-Za-z0-9_,]+)")
-                : Pattern.compile("(?<dir>.*):(?<group>[A-Za-z0-9_,]+)");
+                ? Pattern.compile("(?<dir>|[^A-Za-z?]|[^?]{2,}):(?<group>[A-Za-z0-9_,]+)")
+                : Pattern.compile("(?<dir>[^?]*):(?<group>[A-Za-z0-9_,]+)");
```
sormuras, line 112: Good improvement!

sormuras (changes requested):
Thanks for implementing this feature, Jorn, with updates to the FAQ and tests.

Nit: please delete unused import statements and add a line to the `CHANGELOG.md` file

sbrannen:
> If `foo` has parameters, then for JUnit tests they have to be specified as a comma separated list of binary names.

For array types, I believe that canonical names are also supported -- for example, `int[][]`, `java.lang.String[]`, etc.

JornVernee:
> > If `foo` has parameters, then for JUnit tests they have to be specified as a comma separated list of binary names.
> 
> For array types, I believe that canonical names are also supported -- for example, `int[][]`, `java.lang.String[]`, etc., which might prove beneficial if any of those queries are hand crafted.

Yes, that's a good point. I think the main user of this will be the intellij plugin, which will automatically derive the query string, but it wouldn't hurt to mention this in the FAQ. I'll try to see if there's a JUnit doc section that describes the supported format, and we could just link to that.

JornVernee:
I noticed that there's also a `nested-class` and `nested-method` selector. Is there a reason for keeping these separate? Or is it fine to use the `class` and `method` selectors as well, which work as long as you use the binary name of the nested class.

sormuras:
> I noticed that there's also a nested-class and nested-method selector. Is there a reason for keeping these separate? Or is it fine to use the class and method selectors as well, which work as long as you use the binary name of the nested class?

Binary names for classes and nested classes are fine. These two nested variants can be used for "lazy load" scenarios. See for example the API documentation of `NestedMethodSelector`
> A [DiscoverySelector](https://junit.org/junit5/docs/current/api/org.junit.platform.engine/org/junit/platform/engine/DiscoverySelector.html) that selects a nested [Method](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/reflect/Method.html) or a combination of enclosing classes names, class name, method name, and parameter types so that [TestEngines](https://junit.org/junit5/docs/current/api/org.junit.platform.engine/org/junit/platform/engine/TestEngine.html) can discover tests or containers based on methods.
If a Java [Method](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/lang/reflect/Method.html) is provided, the selector will return that [method](https://junit.org/junit5/docs/current/api/org.junit.platform.engine/org/junit/platform/engine/discovery/NestedMethodSelector.html#getMethod()) and its method name, class name, enclosing classes names, and parameter types accordingly. If class names or method names are provided, this selector will only attempt to lazily load a class or method if [getEnclosingClasses()](https://junit.org/junit5/docs/current/api/org.junit.platform.engine/org/junit/platform/engine/discovery/NestedMethodSelector.html#getEnclosingClasses()), [getNestedClass()](https://junit.org/junit5/docs/current/api/org.junit.platform.engine/org/junit/platform/engine/discovery/NestedMethodSelector.html#getNestedClass()), [getMethod()](https://junit.org/junit5/docs/current/api/org.junit.platform.engine/org/junit/platform/engine/discovery/NestedMethodSelector.html#getMethod()), or [getParameterTypes()](https://junit.org/junit5/docs/current/api/org.junit.platform.engine/org/junit/platform/engine/discovery/NestedMethodSelector.html#getParameterTypes()) is invoked.

JornVernee:
I've updated the FAQ to link to the doc of `DiscoverySelectors#selectMethod(java.lang.String)`

sormuras (approved):
Looks good to me.

jaikiran (approved):
Hello Jorn, these changes look reasonable to me.

Several of these files will need a copyright year update before integrating.

sormuras (approved):

jaikiran (approved):

PR #242, opened 2025-01-14
7903931: Add "test.thread.factory" property to filter tests based on test thread factory value

lmesnik:
jtreg support test thread factory parameter that allows to execute tests by running main in virtual thread.

Currently, all incompatible tests are problemlisted. However after security manager removal the number of such tests is small.

So is it makes sense to add "jtreg.ttf" property to filter tests based on test thread factory value to allow filter out incompatible tests.

The only jtreg knows if -testThreadFactory is set so it should set corresponding property

---------
### Progress

### Issue
 * [CODETOOLS-7903931](https://bugs.openjdk.org/browse/CODETOOLS-7903931): Add "test.thread.factory" property to filter tests based on test thread factory value (**Enhancement** - P4)

### Reviewers
 * [Iris Clark](https://openjdk.org/census#iris) (@irisclark - **Reviewer**)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/242/head:pull/242` \
`$ git checkout pull/242`

`$ git checkout pull/242` \
`$ git pull https://git.openjdk.org/jtreg.git pull/242/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 242`

`$ git pr show -t 242`

</details>

<a href="https://git.openjdk.org/jtreg/pull/242.diff">https://git.openjdk.org/jtreg/pull/242.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/242#issuecomment-2588906728)
</details>

### DIFF
--- a/src/share/classes/com/sun/javatest/regtest/config/RegressionContext.java
+++ b/src/share/classes/com/sun/javatest/regtest/config/RegressionContext.java
@@ -104,7 +104,7 @@ public class RegressionContext implements Expr.Context {
         String testThreadFactory = (params == null || params.getTestThreadFactory() == null)
                 ? "null"
                 : params.getTestThreadFactory();
-        values.put("jtreg.ttf", testThreadFactory);
+        values.put("test.thread.factory", testThreadFactory);
 
         processVMOptions((params == null) ? Collections.<String>emptyList() : params.getTestVMJavaOptions());
     }
--- a/src/share/classes/com/sun/javatest/regtest/config/RegressionContext.java
+++ b/src/share/classes/com/sun/javatest/regtest/config/RegressionContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2014, 2018, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 2014, 2025, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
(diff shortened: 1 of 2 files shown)


sormuras:
The namespace `jtreg.` is not used in `RegressionContext` yet, but it makes sense to introduce it here.

Is `ttf` a good abbreviation for `test thread factory` though? Would `jtreg.testThreadFactory` be too long in filter expressions?

lmesnik:
> The namespace `jtreg.` is not used in `RegressionContext` yet, but it makes sense to introduce it here.
> 
> Is `ttf` a good abbreviation for `test thread factory` though? Would `jtreg.testThreadFactory` be too long in filter expressions?

I think `jtreg.ttf` is good enough, but might be becasue I know the context. If you feel that for all jtreg users `jtreg.testThreadFactory`, I am fine with this.

jonathan-gibbons:
> > The namespace `jtreg.` is not used in `RegressionContext` yet, but it makes sense to introduce it here.
> > Is `ttf` a good abbreviation for `test thread factory` though? Would `jtreg.testThreadFactory` be too long in filter expressions?
> 
> I think `jtreg.ttf` is good enough, but might be becasue I know the context. If you feel that for all jtreg users `jtreg.testThreadFactory`, I am fine with this.

For the long term, clarity is better than brevity.

sormuras:
What about re-using the `test.` namespace used by `jtreg` for [system properties](https://openjdk.org/jtreg/tag-spec.html#testvars)? Meaning, we'd call this property introduced here:
- `test.threadFactory`

lmesnik:
I'm fine with using test namespace.
Might be
- test.thread.factory
then?

It is the name of property used by test to check thread factory inside of test.
Leonid

sormuras:
We have winner! `test.thread.factory`

src/share/classes/com/sun/javatest/regtest/config/RegressionContext.java
```diff
@@ -101,6 +101,11 @@ static RegressionContext getDefault() {
         values.put("os.maxMemory", String.valueOf(os.maxMemory));
         values.put("os.maxSwap", String.valueOf(os.maxSwap));
 
+        String testThreadFactory = (params == null || params.getTestThreadFactory() == null)
+                ? "null"
```
sormuras, line 105: Having `null` as the value for "no test thread factory available" is right choice regarding easier handling on the use-site?

Or would skipping the `values.put()` call be a better default? 🤔

lmesnik: I haven't checked but it should be considered as a "invalid name" if we don't put any value.
Also, want to be consistent with other properties, that just have some default value, often "null".

sormuras (approved):

irisclark (approved):

PR #248, opened 2025-02-24
7903953: Use scratch directory for JUnit's TempDir annotation

sormuras:
Please review this change to use `jtreg`'s scratch directory for JUnit's `@TempDir` annotation.

Prior to this change, the standard implementation of JUnit Jupiter using the system default temporary directory is active, effectively decoupling the location of "interesting on failure" files from `jtreg`:
> When `jtreg` executes a test, the current directory for the test is set to a scratch directory so that the test can easily write any temporary files.

Find more details about `jtreg`'s scratch directory at https://openjdk.org/jtreg/faq.html#scratch-directory

In addition to use the current working directory for `@TempDir` annotated paths, this change also switches off JUnit's built-in cleanup of temporary files - leaving it to `jtreg` to care about it, see for example the help message for  the `-retain` command-line option:

```
    -retain | -retain:<pass,fail,error,all,file-pattern>,...
                    Specify files to be retained after each test completes
                    executing. If -retain is not specified, only the files from
                    the last test executed will be retained. If -retain is
                    specified with no argument, all files will be retained.
                    Otherwise, the files may be described by one or more of the
                    following values:
        none            Do not retain any of the files generated by each test
        pass            Retain files generated by tests that pass
        fail            Retain files generated by tests that fail
        error           Retain files generated by tests that caused an error
        all             Retain all files generated by each test
        file-pattern    Retain files that match a specific filename. The name
                        may contain '*' to match any sequence of characters. For
                        example, result.* or *.err.
```

---------
### Progress

### Issue
 * [CODETOOLS-7903953](https://bugs.openjdk.org/browse/CODETOOLS-7903953): Use scratch directory for JUnit's TempDir annotation (**Enhancement** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/248/head:pull/248` \
`$ git checkout pull/248`

`$ git checkout pull/248` \
`$ git pull https://git.openjdk.org/jtreg.git pull/248/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 248`

`$ git pr show -t 248`

</details>

<a href="https://git.openjdk.org/jtreg/pull/248.diff">https://git.openjdk.org/jtreg/pull/248.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/248#issuecomment-2677711781)
</details>

### DIFF
--- a/test/junitTrace/JupiterTempDir.java
+++ b/test/junitTrace/JupiterTempDir.java
@@ -21,13 +21,14 @@
  * questions.
  */
 
-import org.junit.jupiter.api.Assertions;
-import org.junit.jupiter.api.Test;
-import org.junit.jupiter.api.condition.EnabledIfSystemProperty;
-import org.junit.jupiter.api.io.TempDir;
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 import java.nio.file.Files;
 import java.nio.file.Path;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
 
 /*
  * @test
@@ -36,13 +37,15 @@ import java.nio.file.Path;
  */
 class JupiterTempDir {
     @Test
-    @EnabledIfSystemProperty(named = "test.name", matches = "JupiterTempDir.java")
-    void test(@TempDir Path temporary) throws Exception {
+    void currentWorkingDirectoryIsParentOfTemporaryDirectory(@TempDir Path temporary) {
+        assumeTrue(System.getProperty("test.file") != null, "jtreg not running");
+
+        assertTrue(Files.isDirectory(temporary), "temporary directory expected");
+
         var currentWorkingDirectory = Path.of("");
         var expected = currentWorkingDirectory.toAbsolutePath().toString();
         var actual = temporary.getParent().toAbsolutePath().toString();
-        Assertions.assertEquals(expected, actual, "Unexpected root for temporary files");
 
-        Files.createTempFile(temporary, "temp-", ".temp");
+        assertEquals(expected, actual, "unexpected parent directory for temporary files");
     }
 }
--- a/test/junitTrace/JupiterTempDir.java
+++ b/test/junitTrace/JupiterTempDir.java
@@ -38,7 +38,9 @@ import org.junit.jupiter.api.io.TempDir;
 class JupiterTempDir {
     @Test
     void currentWorkingDirectoryIsParentOfTemporaryDirectory(@TempDir Path temporary) {
-        assumeTrue(System.getProperty("test.file") != null, "jtreg not running");
+        // Verify that this test was launched through jtreg launcher, which sets the `test.file`
+        // system property. Find a list of them at https://openjdk.org/jtreg/tag-spec.html#testvars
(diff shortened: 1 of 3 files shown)


test/junitTrace/JupiterTempDir.java
```diff
@@ -0,0 +1,48 @@
...
+import java.nio.file.Path;
+
+/*
+ * @test
+ * @bug 7903953
+ * @run junit JupiterTempDir
+ */
+class JupiterTempDir {
+    @Test
+    @EnabledIfSystemProperty(named = "test.name", matches = "JupiterTempDir.java")
```
jaikiran, line 39: Hello Christian, why is this needed in this test?

test/junitTrace/JupiterTempDir.java
```diff
@@ -0,0 +1,48 @@
...
+    void test(@TempDir Path temporary) throws Exception {
+        var currentWorkingDirectory = Path.of("");
+        var expected = currentWorkingDirectory.toAbsolutePath().toString();
+        var actual = temporary.getParent().toAbsolutePath().toString();
+        Assertions.assertEquals(expected, actual, "Unexpected root for temporary files");
```
jaikiran, line 44: Would `Files.isSameFile(currentWorkingDirectory, temporary.getParent())` be more appropriate here?

test/junitTrace/JupiterTempDir.java
sormuras: The assertion(s) in the test only make sense, when `jtreg` runs the test via its `JUnitRunner`. The `test.name` system property is an indicator that `jtreg` runs the test, I could switch to another one documented here: https://openjdk.org/jtreg/tag-spec.html#testvars

test/junitTrace/JupiterTempDir.java
sormuras: Not sure if this would hold true for every case, especially for some fancy (sym-)linked directories?

I think the `String`-based comparison is good-enough here, and has a nice side-effect producing a human-readable error messages when the assertion fails.

test/junitTrace/JupiterTempDir.java
sormuras: Perhaps it would be better to replace this non-obvious declarative guard with a programmatic `Assumptions....` statement within the test method. 🤔

jaikiran: > The assertion(s) in the test only make sense, when jtreg runs the test via its JUnitRunner.

Since this is a self test in the jtreg project, would it be the case that a test marked as `@run junit` will always be run by the `JUnitRunner`? So even the assumption I think is probably isn't needed.

sormuras: When running this self-test without jtreg in play, for example via a direct launch within an IDE(A), it would fail hard. With the assumption in place, developers see:

```
org.opentest4j.TestAbortedException: Assumption failed: jtreg not running

	at org.junit.jupiter.api.Assumptions.throwAssumptionFailed(Assumptions.java:316)
	at org.junit.jupiter.api.Assumptions.assumeTrue(Assumptions.java:115)
	at JupiterTempDir.currentWorkingDirectoryIsParentOfTemporary(JupiterTempDir.java:44)
```

Adding a comment, as suggested in another channel, will help future folk though.

jaikiran (approved):
Thank you Christian for the updates. This looks good to me.

PR #250, opened 2025-03-05
7903963: Fix "Run Configuration" Window Freeze Issue

asilichenko:
# When the Problem Occurs

The JTReg "Run Configuration" window freezes when the user selects a JRE that is not present in the project's SDK list.

# Why the Problem Occurs

There is a method `applyEditorTo` in the `JTRegConfigurable.java` class, which is called repeatedly while the "Run Configuration" window is open. Within this method, the method `jrePathEditor.getJrePathOrName()` is invoked. As a result, when this method is called for a JRE path not listed in IDEA, it attempts to determine the Java version. If the user has selected a raw JDK image, a Java process is launched, and the version is obtained from the process output (see [org.jetbrains.jps.model.java.impl.JdkVersionDetectorImpl#detectJdkVersionInfo](https://github.com/JetBrains/intellij-community/blob/baba77ecf93f44a2ccb34e5b206e03c60817d511/jps/model-impl/src/org/jetbrains/jps/model/java/impl/JdkVersionDetectorImpl.java#L40)). 

This call is too resource-intensive to be executed on the UI thread, causing the window to freeze.

# Steps to reproduce

1. Open a JDK project.  
2. Open the "Project Structure" window.  
3. Navigate to "Platform Settings → SDKs".
4. Remove the JDK that is used as the JRE for tests, or create a copy of this JDK on your file system.  
5. Create a new "JTReg Run Configuration".  
6. In the "JRE" field, select a JDK path that is not present in the Project SDK list (either the removed JDK or the copied one).  

## Actual result
- The window starts freezing, even after saving and reopening it.

# Implemented Changes

## JTRegConfigurable.java
- Fix Javadoc for the class.
- Remove unused class field: `jtregDir`.
- Add Javadoc to the methods: `createUIComponents`, `applyEditorTo`, `resetEditorFrom`
- Fix window freeze caused by `jrePathEditor` in the method `applyEditorTo`.
- Add method `applyEditorJrePathTo` to optimize data retrieval from `jrePathEditor`.

Since I implemented changes in this file, I also took the opportunity to clean up the code a bit.

---------
### Progress

### Issue
 * [CODETOOLS-7903963](https://bugs.openjdk.org/browse/CODETOOLS-7903963): Fix "Run Configuration" Window Freeze Issue (**Bug** - P4)

### Reviewers
 * [Jorn Vernee](https://openjdk.org/census#jvernee) (@JornVernee - Committer)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/250/head:pull/250` \
`$ git checkout pull/250`

`$ git checkout pull/250` \
`$ git pull https://git.openjdk.org/jtreg.git pull/250/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 250`

`$ git pr show -t 250`

</details>

<a href="https://git.openjdk.org/jtreg/pull/250.diff">https://git.openjdk.org/jtreg/pull/250.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/250#issuecomment-2706993787)
</details>

### DIFF
--- a/plugins/idea/src/main/java/com/oracle/plugin/jtreg/configuration/ui/JTRegConfigurable.java
+++ b/plugins/idea/src/main/java/com/oracle/plugin/jtreg/configuration/ui/JTRegConfigurable.java
@@ -30,22 +30,26 @@ import com.intellij.execution.ui.JrePathEditor;
 import com.intellij.openapi.fileChooser.FileChooserDescriptorFactory;
 import com.intellij.openapi.options.SettingsEditor;
 import com.intellij.openapi.project.Project;
+import com.intellij.openapi.projectRoots.JdkUtil;
+import com.intellij.openapi.ui.ComboBox;
 import com.intellij.openapi.ui.TextFieldWithBrowseButton;
 import com.intellij.openapi.util.io.FileUtil;
+import com.intellij.ui.components.JBTextField;
 import com.oracle.plugin.jtreg.configuration.JTRegConfiguration;
 import org.jetbrains.annotations.NotNull;
 
 import javax.swing.*;
 import java.awt.event.ActionEvent;
 import java.awt.event.ActionListener;
+import java.util.Objects;
+import java.util.stream.IntStream;
 
 /**
- * This class models the dialog associated with the (project-wide) jtreg tool settings.
+ * This class models the "Run/Debug Configurations" dialog for the JTReg tool settings.
  */
 public class JTRegConfigurable<T extends JTRegConfiguration> extends SettingsEditor<T> {
     private JTextField jtregQuery;
     private JTextField jtregOptions;
-    private TextFieldWithBrowseButton jtregDir;
     private TextFieldWithBrowseButton workDirectory;
     private JrePathEditor jrePathEditor;
     private JPanel mainPane;
@@ -63,10 +67,12 @@ public class JTRegConfigurable<T extends JTRegConfiguration> extends SettingsEdi
         directoryRadioButton.addActionListener(listener);
     }
 
+    /**
+     * Gets called for each JTReg configuration entry
+     * when it appears in the "Run/Debug Configurations" window.
+     */
     private void createUIComponents() {
         jrePathEditor = new JrePathEditor(DefaultJreSelector.projectSdk(project));
-        jtregDir = new TextFieldWithBrowseButton();
-        jtregDir.addBrowseFolderListener("Directory with Strategies", null, project, FileChooserDescriptorFactory.createSingleFolderDescriptor());
         workDirectory = new TextFieldWithBrowseButton();
         workDirectory.addBrowseFolderListener("Directory with Strategies", null, project, FileChooserDescriptorFactory.createSingleFolderDescriptor());
         file = new TextFieldWithBrowseButton();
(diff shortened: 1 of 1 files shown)


JornVernee:
> Hi, I’ve created this pull request, but I don’t have the necessary permissions to create an issue for it. Could someone please create an issue to track this change? I’m not sure who to contact directly, so any help would be appreciated.

Typically your best bet is to send a message to one of the mailing lists: https://mail.openjdk.org/mailman/listinfo For this project, the right list would be ide-support-dev@openjdk.org

For now I'll make a JBS issue for you.

JornVernee:
I was able to reproduce the issue. I noticed that is was particularly noticeable when I select the JRE field, and then select another field. The UI seems to freeze for 1-2 seconds.

I've created the issue: https://bugs.openjdk.org/browse/CODETOOLS-7903963 Please change the PR title to:

```
7903963: Fix "Run Configuration" Window Freeze Issue
```

plugins/idea/src/main/java/com/oracle/plugin/jtreg/configuration/ui/JTRegConfigurable.java
```diff
@@ -97,6 +108,63 @@ public void applyEditorTo(final JTRegConfiguration configuration) {
...
+        final boolean alternativeJREPathEnabled;
+        if (inList) { // safe to get item from JRE path editor
+            alternativeJrePath = jrePathEditor.getJrePathOrName();
+            alternativeJREPathEnabled = jrePathEditor.isAlternativeJreSelected();
+        } else { // JRE path editor would be time-consuming
+            alternativeJrePath = FileUtil.toSystemIndependentName(jrePathOrNameText); // The value here should always be the path, not the name.
+            alternativeJREPathEnabled = true;
+            if (JdkUtil.checkForJdk(alternativeJrePath)) {
+                // If the path is a valid JDK, add it to the ComboBox list.
+                jrePathEditor.setPathOrName(alternativeJrePath, true);
```
JornVernee, line 156: I'd like to better understand the effect that adding the path to the list is intended to have. I think the main point here is that the next time `applyEditorTo` is called, the path will be in the list, and we will take the `if` branch instead, which at that point is faster?

The custom JRE path doesn't seem to persist if I close and re-open the run configuration dialogue.

JornVernee:
I've verified that the fix works on my machine as well.

plugins/idea/src/main/java/com/oracle/plugin/jtreg/configuration/ui/JTRegConfigurable.java
```diff
@@ -97,6 +108,63 @@ public void applyEditorTo(final JTRegConfiguration configuration) {
...
+
+        final String jrePathOrNameText = jreTextField.getText().trim();
+        final boolean inList = IntStream.range(0, jreComboBox.getItemCount())
+                .mapToObj(jreComboBox::getItemAt)
+                .map(JrePathEditor.JreComboBoxItem::getPresentableText)
+                .filter(Objects::nonNull)
+                .anyMatch(pathOrName -> pathOrName.equals(jrePathOrNameText));
+
+        final String alternativeJrePath;
+        final boolean alternativeJREPathEnabled;
```
JornVernee, line 147: I've discussed with my colleague, and in general we'd like to not have `final` on local variables. I know some parts of the plugin don't follow that rule, but we shouldn't add more of it.

Please remove `final` from local variables.

plugins/idea/src/main/java/com/oracle/plugin/jtreg/configuration/ui/JTRegConfigurable.java
asilichenko: Yes, this is so that the next time `applyEditorTo` is called, we can safely call `jrePathEditor.getJrePathOrName()` and `jrePathEditor.isAlternativeJreSelected()` (the `if` case).

The custom JRE path does indeed only exist in the list while the window is open.
I added this code to ensure that in the general case, the original `getJrePathOrName` and `isAlternativeJreSelected` calls are used.

plugins/idea/src/main/java/com/oracle/plugin/jtreg/configuration/ui/JTRegConfigurable.java
asilichenko: done

JornVernee (approved):

sormuras (approved):

PR #253, opened 2025-03-07
7903961: Make default timeout configurable

sormuras:
Please review this change replacing the hard-coded `120` seconds default timeout for test actions with a configurable solution parsing a custom default value from `TEST.ROOT` and `TEST.properties` files:

```
timeout.default.seconds=360
```

Especially in combination with "JUnit.dirs" and "TestNG.dirs", where all actions directives found in a test file are ignored, is helpful.

---------
### Progress

### Issue
 * [CODETOOLS-7903961](https://bugs.openjdk.org/browse/CODETOOLS-7903961): Make default timeout configurable (**Enhancement** - P3)

### Reviewers
 * [Jaikiran Pai](https://openjdk.org/census#jpai) (@jaikiran - **Reviewer**) ⚠️ Review applies to [b9c04050](https://git.openjdk.org/jtreg/pull/253/files/b9c040509a577125ce180cf454920d82ae72be14)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/253/head:pull/253` \
`$ git checkout pull/253`

`$ git checkout pull/253` \
`$ git pull https://git.openjdk.org/jtreg.git pull/253/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 253`

`$ git pr show -t 253`

</details>

<a href="https://git.openjdk.org/jtreg/pull/253.diff">https://git.openjdk.org/jtreg/pull/253.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/253#issuecomment-2710104659)
</details>


src/share/classes/com/sun/javatest/regtest/exec/RegressionScript.java
```diff
@@ -1123,7 +1128,7 @@ Path getJavacProg() {
 
     Map<String, String> getTestProperties()  {
         // initialize the properties with standard properties common to all tests
-        Map<String, String> p = new LinkedHashMap<>(params.getBasicTestProperties());
+        Map<String, String> p = new TreeMap<>(params.getBasicTestProperties());
```
sormuras, line 1131: This change sorts all basic test properties by their name; keeping the insertion order was never a promise.

src/share/classes/com/sun/javatest/regtest/config/TestProperties.java
```diff
@@ -270,6 +276,9 @@ class Entry {
                     // add the maxOutputSize for result content
                     maxOutputSize = getInt("maxOutputSize", -1);
 
+                    // add the default test timeout value in seconds
+                    defaultTimeout = Duration.ofSeconds(getInt("timeout.default.seconds", 120));
```
jaikiran, line 280: Hello Christian, the `TEST.ROOT` (and `TEST.properties`) appear to follow the camel case style property names for most of the configurations https://openjdk.org/jtreg/tag-spec.html#TEST_ROOT. For example, `defaultExecMode`, `maxOutputSize` and so on. To be consistent with that, maybe we should expect this to be `defaultTimeoutSeconds` instead of `timeout.default.seconds`?

src/share/classes/com/sun/javatest/regtest/exec/RegressionScript.java
jaikiran: Does the new proposed feature of configuring a default timeout continue to work without this change to using a `TreeMap`? If yes, then I think it would be good to separate this change out from this PR.

src/share/classes/com/sun/javatest/regtest/exec/RegressionScript.java
```diff
@@ -1168,6 +1173,10 @@ Map<String, String> getTestProperties()  {
         if (disablePreview()) {
             p.put("test.enable.preview", "false");
         }
+        if (isTimeoutsEnabled()) {
+            long seconds = properties.getDefaultTimeout(getTestDescription().getFile()).toSeconds();
+            p.put("test.timeout.default.seconds", Long.toString(seconds));
```
jaikiran, line 1178: Is this the properties that get advertised to the application's test code? I'm guessing it is because I see that the `tag-spec` documentation has been updated to mention that this property will be made available to the test code. 

Is there a reason why the test code would be interested in this default value? As far as I can see we don't expose the final evaluated timeout value nor do we expose any other default configuration values (like the `defaultExecMode`) as properties to the test code. So it might be better not to expose this at least until it is really necessary.

src/share/classes/com/sun/javatest/regtest/exec/RegressionScript.java
```diff
@@ -1342,5 +1351,14 @@ private static String getVersion() {
     private Modules modules;
     private ScratchDirectory scratchDirectory;
     Locations locations;
+    private final TestProperties properties;
+
+    private RegressionScript() {
+        this(null);
```
jaikiran, line 1357: Does this constructor get used anywhere after this change? If yes, then we might have to add a null check at the call site where we are using the `properties` field to find the default timeout configuration value, like here https://github.com/openjdk/jtreg/pull/253/files#diff-19a61ca7f0fe24f891bee1b6b95f4ab2f740636458c9b20acff20c052aed9223R539

test/junitTrace/JupiterTimeoutTest.java
```diff
@@ -0,0 +1,40 @@
...
+/*
+ * @test
+ * @bug 7903961
+ * @run junit JupiterTimeoutTest
+ */
+class JupiterTimeoutTest {
+    @Test
+    void test() {
+        String actual = System.getProperty("test.timeout.default.seconds", "-1");
+        assertEquals("123", actual, "unexpected default value for timeout seconds");
```
jaikiran, line 38: It would be good to also include a test which overrides the default value in the test definition and verify that the overridden value is used. After all, that's the usage which prompted this change.

src/share/classes/com/sun/javatest/regtest/config/TestProperties.java
sormuras: It's quite a mixture already. There's also `lib.build` and `external.lib.roots` as counter examples.

The idea was to synchronize the key name with the system property exposed to tests in `RegressionScript`.

src/share/classes/com/sun/javatest/regtest/exec/RegressionScript.java
jaikiran: Hello Christian, I had forgotten that the test timeout factor gets exposed to test code as a property. Having said that, the test timeout factor is the final determined value unlike this proposed default timeout property which is a default value and may not even be the actual timeout that's ultimately used by jtreg.

src/share/classes/com/sun/javatest/regtest/config/TestProperties.java
jaikiran: I don't have a strong opinion about this naming in the `TEST.ROOT`/`TEST.properties`, because like you state it's already a mix of naming styles.

src/share/classes/com/sun/javatest/regtest/exec/RegressionScript.java
sormuras: Good point. Reverting this change.

src/share/classes/com/sun/javatest/regtest/exec/RegressionScript.java
sormuras: As discussed offline, this PR won't export that property. It can be done in a follow-up work.

src/share/classes/com/sun/javatest/regtest/exec/RegressionScript.java
```diff
@@ -1342,5 +1346,10 @@ private static String getVersion() {
     private Modules modules;
     private ScratchDirectory scratchDirectory;
     Locations locations;
+    private final TestProperties properties;
+
+    public RegressionScript(TestProperties properties) {
```
jaikiran, line 1351: I did not notice this in the previous round, sorry, but this constructor can be `private` like the other one that previously was.

sormuras: No, it has to remain `public` because it is used by `RegressionTestSuite` in the `config` package.

jaikiran: You are right, I missed that.

jaikiran (approved):
These changes look OK to me.

PR #259, opened 2025-04-14
Default to no error if existing, make parent directories as needed

sormuras:
Prior this change, most usages of the `$(MKDIR)` variable in jtreg's self-tests had to add `-p` in order to not fail with an "error if existing, make parent directories as needed". Many usages are lacking the flag and in concurrent test scenarios, this leads to flaky tests.

This PR appends `-p` to each expansion of `$(MKDIR)`. Hard-coded `-p` will be removed; stray/duplicate `-p` flag don't seem to be a problem for `mkdir`'s option parser. I didn't see any "-p specifed more than once" warnings.

### DIFF
--- a/make/Defs.gmk
+++ b/make/Defs.gmk
@@ -152,7 +152,7 @@ GREP 	:= $(shell if [ -r /bin/grep ]; then echo /bin/grep ; else echo /usr/bin/g
 endif
 LN	= /bin/ln
 LS	= /bin/ls
-MKDIR 	= /bin/mkdir
+MKDIR 	= /bin/mkdir -p
 MV 	= /bin/mv
 PANDOC  := $(shell if [ -r /usr/bin/pandoc ]; then \
 		echo /usr/bin/pandoc ; \


test/jdkOptsTest/JDKOptsTest.gmk
```diff
@@ -29,14 +29,14 @@ $(BUILDTESTDIR)/JDKOptsTest.ok: \
...
 	    $(JTREG_IMAGEDIR)/lib/jtreg.jar \
 	    $(JTREG_IMAGEDIR)/bin/jtreg
-	$(MKDIR) -p $(@:%.ok=%)/classes
+	$(MKDIR) $(@:%.ok=%)/classes
 	$(JDKJAVAC) \
 		-d $(@:%.ok=%)/classes \
 		-cp $(JTREG_IMAGEDIR)/lib/jtreg.jar \
 		-Xlint -Werror \
 		-encoding ASCII $(TESTDIR)/jdkOptsTest/JDKOptsTest.java
 	$(RM) -rf $(@:%.ok=%)/tmp
```
irisclark, line 38: `$(RM)` is already defined to include the `-rf`, so suggest you remove those options here

irisclark (approved):
I appreciate the cleanup for `$(RM)`.

Thanks.

test/jdkOptsTest/JDKOptsTest.gmk
sormuras: Good catch. I only searched for `$(RM) -f` patterns before. Searching for `$(RM) -rf` hit two locations, both addressed in https://github.com/openjdk/jtreg/pull/259/commits/167328701a7f253b0eadbfe638545063fb217e5e

sormuras: By the way, no hits for `$(RM) -fr` ... 😉

irisclark (approved):

PR #260, opened 2025-04-15
7903990: IDEA plugin: Itemize junit test results

JornVernee:
Itemize the results of jtreg tests containing junit sections. Each junit section will have a node in the test result tree, with each nested test class and test method having their own node as well. This also makes it possible to re-run just a single junit test (or a single iteration of a parameterized test), by right-clicking the corresponding item in the test result UI. See the below image as an example:

image

Some implementation notes: jtharness/jtreg don't currently report results of individual junit tests to the test listener. There are currently 2 barriers to this that I can see: 

1. the JUnitRunner communicates the test results back to the parent jtreg process by writing a summary to stderr, which the parent process than parses. The parent process does not parse the results of the individual tests. In order to have more 'live feedback' from the junit runner, one idea is to use a socket connection between the parent jtreg process and junit runner. Where instead of writing the test results to stderr, the junit runner would call back to the parent process over this socket connection to notify it when a test started/finished. The jtreg parent process can then in turn immediately notify an test listener. Alternatively, we could perhaps use (existing) junit JFR events that are streamed back to the parent process. 
2. jtreg does not report nested events of a test (such as junit tests) back to the harness observer that the plugin uses to listen for test events. In jtharness, a 'test' corresponds directly to a jtreg test, and there seems to be no support to attach additional nested events to a 'test', or for the JTRegTestListener to register a callback to listen for more nested events.

Rather than solving these two issues, for now I've opted for a simpler approach. When a jtreg test finishes and the test listener is notified, we look for sections in the test result titled 'junit', and parse their stderr stream to find the tests results that jtreg's JUnitRunner prints to stderr. This works relatively well, with the caveat being that, if jtharness truncates the stderr output, some tests that ran will not be reported in the UI in an itemized way (but the test will still fail, and report the failure under the 'jtreg' test, accompanied by its jtr file). Given how useful this feature seems (and has been during testing), I think this is a reasonable tradeoff.

---------
### Progress

### Issue
 * [CODETOOLS-7903990](https://bugs.openjdk.org/browse/CODETOOLS-7903990): IDEA plugin: Itemize junit test results (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/260/head:pull/260` \
`$ git checkout pull/260`

`$ git checkout pull/260` \
`$ git pull https://git.openjdk.org/jtreg.git pull/260/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 260`

`$ git pr show -t 260`

</details>

<a href="https://git.openjdk.org/jtreg/pull/260.diff">https://git.openjdk.org/jtreg/pull/260.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/260#issuecomment-2812702359)
</details>


JornVernee:
Another caveat of this feature I've run into, or rather, what seems to be a limitation imposed by IntelliJ: in order to show the right-click dialogue to re-run a test case, we need to look up the location of the test class/method, which then gets passed back to the plugin later to create the run configuration. This involves at least looking up a class using IntelliJ's builtin lookup functionality.

The issue is: when we look up the name of the test class, the file it is in needs to be in an active IntelliJ test source root. The plugin will normally automatically set up the test source root when opening a test file, but if you forget to do this before running the test, the test class can not be found, and instead of the dialogue from the screen shot, a dialogue with the text 'Nothing here' will be shown instead.

image

The worse thing is that this lookup result seems to be cached inside the IDE, so even if you add the test source root after the seeing this, the IDE will keep showing the 'Nothing here' dialogue. The only way I've found to resolve that issue is to restart the IDE.

JornVernee:
There's one more issue I've discovered: when running tests concurrently (using e.g. `-conc:auto`), the tests will get nested into each other:

image

I've made several attempts to work around this issue, for instance by assigning a different 'flow id' to each test (as described [here](https://www.jetbrains.com/help/teamcity/service-messages.html#Message+FlowId)), but this parameter seems to be ignored by intellij (which I've found out by debugging the IDE code itself) (see e.g. [here](https://github.com/JetBrains/intellij-community/blob/40473f1b030aa26244a8051df8d8a90fdbafbec6/platform/smRunner/src/com/intellij/execution/testframework/sm/runner/GeneralToSMTRunnerEventsConvertor.java#L139) where the code only has a single flowId-insensitive stack of test suites.) It seems to be possible to create the entire event tree up front, and this seems to be what the junit runner does, but unlike the junit runner, we don't know the structure of the event tree before we start running tests, so this isn't really an option for us.

I don't think concurrent test running ever really  worked in the plugin. With the old code we don't get the weird nesting, but each new test started would immediately end an already running test. For now I think the best solution to make concurrent tests work with the new setup, is to only report the 'test suite started' message from `finishedTest`, under a lock. This gets us a correct event tree, with the downside that tests only start appearing in the UI once they are finished. Seems like a decent tradeoff?

sormuras:
> [...]  This gets us a correct event tree, with the downside that tests only start appearing in the UI once they are finished. Seems like a decent tradeoff?

Yes. Better show correct information then incorrect one.

Is it possible to switch to that more defensive strategy only if multi threading is activated?

JornVernee:
> Is it possible to switch to that more defensive strategy only if multi threading is activated?

Yes, that seems to be possible. The test listener has access to that information.

JornVernee:
This fell off my radar for a bit, but I've now fixed concurrent test output. When a test starts, it tries to grab a writer lock, and tries to write out the test start event immediately. but only 1 test is allowed to do this at a time, depending on which test grabbed the lock. For tests that failed to grab the lock, the start test event is deferred until the test is finished. This will allow multiple tests to run concurrently, but only 1 test can output test events at a time.

sormuras (approved):
Looks good to me.

JornVernee:
I gave this a last look, and noticed I had accidentally included some spurious lines in the last commit, and there was a missing label in one of the loops. I've addressed both in https://github.com/openjdk/jtreg/pull/260/commits/603de523c4e34597698330c0945dfaa84d36e80d

sormuras (approved):
Re-approved.

PR #262, opened 2025-05-06
7904004: Run jtreg GHA workflows for FreeBSD

snake66:
GitHub Actions does not natively support FreeBSD, so we use the vmactions/freebsd-vm action as a base to run the tests in a VM under an Ubuntu runner.

This work is sponsored by The FreeBSD Foundation.

---------
### Progress

### Issue
 * [CODETOOLS-7904004](https://bugs.openjdk.org/browse/CODETOOLS-7904004): Run jtreg GHA workflows for FreeBSD (**Enhancement** - P4)

### Reviewers
 * [Magnus Ihse Bursie](https://openjdk.org/census#ihse) (@magicus - Author)

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/262/head:pull/262` \
`$ git checkout pull/262`

`$ git checkout pull/262` \
`$ git pull https://git.openjdk.org/jtreg.git pull/262/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 262`

`$ git pr show -t 262`

</details>

<a href="https://git.openjdk.org/jtreg/pull/262.diff">https://git.openjdk.org/jtreg/pull/262.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/262#issuecomment-2854473436)
</details>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -70,6 +70,6 @@ jobs:
             sysctl hw.physmem
             sysctl hw.usermem
             pw user add -n action -m
-            su action -c 'git config --global --add safe.directory /home/runner/work/jtreg/jtreg'
+            su action -c 'git config --global --add safe.directory $(pwd)'
             su action -c 'bash make/build.sh --jdk /usr/local/openjdk21'
             su action -c 'bash build/make.sh test'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -70,4 +70,6 @@ jobs:
             sysctl hw.physmem
             sysctl hw.usermem
             pw user add -n action -m
-            su action -c 'git config --global --add safe.directory /home/runner/work/jtreg/jtreg && bash make/build.sh --jdk /usr/local/openjdk21 && bash build/make.sh test'
+            su action -c 'git config --global --add safe.directory /home/runner/work/jtreg/jtreg'
+            su action -c 'bash make/build.sh --jdk /usr/local/openjdk21'
+            su action -c 'bash build/make.sh test'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -44,3 +44,30 @@ jobs:
         HEADLESS: 1
       run: |
         bash make/build.sh --skip-download
+
+  freebsd-x64:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - name: Run tests in FreeBSD
+        id: test
+        uses: vmactions/freebsd-vm@v1
+        with:
+          prepare: |
+            pkg install -y bash curl git gmake gnugrep openjdk21 zip
+            mount -t fdescfs fdesc /dev/fd
+            mount -t procfs proc /proc
+
+          run: |
+            pwd
+            ls -lah
+            whoami
+            env
+            freebsd-version
+            sysctl hw.model
+            sysctl hw.ncpu
+            sysctl hw.physmem
+            sysctl hw.usermem
+            pw user add -n action -m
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -44,3 +44,30 @@ jobs:
...
+            ls -lah
+            whoami
+            env
+            freebsd-version
+            sysctl hw.model
+            sysctl hw.ncpu
+            sysctl hw.physmem
+            sysctl hw.usermem
+            pw user add -n action -m
+            su action -c 'git config --global --add safe.directory /home/runner/work/jtreg/jtreg && bash make/build.sh --jdk /usr/local/openjdk21 && bash build/make.sh test'
```
sormuras, line 73: Could the `git` configuration and two make calls be split into dedicated lines?

snake66: Probably, I'll give it a try!

.github/workflows/test.yml
```diff
@@ -44,3 +44,32 @@ jobs:
...
+            ls -lah
+            whoami
+            env
+            freebsd-version
+            sysctl hw.model
+            sysctl hw.ncpu
+            sysctl hw.physmem
+            sysctl hw.usermem
+            pw user add -n action -m
+            su action -c 'git config --global --add safe.directory /home/runner/work/jtreg/jtreg'
```
sormuras, line 73: Looks much better now, thanks for splitting the line into three calls.

Can the hard-coded `/home/runner/work/jtreg/jtreg` path be replaced with some inline `pwd` evaluation? Would make it more future-proof if parent paths are changed.

.github/workflows/test.yml
```diff
@@ -44,3 +44,32 @@ jobs:
...
+            su action -c 'bash make/build.sh --jdk /usr/local/openjdk21'
```
sormuras, line 74: I assume that the hard-coded `/usr/local/openjdk21` is well-defined and stable for `pkg ... openjdk21 ...` calls. Or is there an environment variable (`JAVA_HOME`?) being set that can be read out and used here?

snake66: While it's technically possible to change the prefix from `/usr/local`, this is the default unless explicitly changed.

There are ways to get the path without hardcoding it, but I'm not sure it's worth the effort here:

```
% JAVAVM_DRYRUN=1 javavm 2>/dev/null |grep JAVA_HOME|cut -d '=' -f 2
/usr/local/openjdk21
```

sormuras: The default is good-enough here, I guess. Otherwise, it would be more elegant to specify a custom location for the installation and then pass it later to subsequent processes.

sormuras:
Running for almost an hour now and repeating `VM is booting` over and over. This does not bode well...

image

snake66:
> Running for almost an hour now and repeating `VM is booting` over and over. This does not bode well...

No, tried restarting it. Not seen this happen before.

magicus:
I think that sounds like a much worse solution. Let's stick to GHA. @sormuras Are there anything else you think needs to be done for this? Otherwise I believe this would be good to go.

sormuras:
> Let's stick to GHA.

Yes.

> Are there anything else you think needs to be done for this? Otherwise I believe this would be good to go.

Except for this "eternal boot loop" issue, no. If that happens too often, we can always disable or remove this new job.

sormuras (approved):
Looks good to me.

Will be observing the "boot loop" as a potential reason to back out.

snake66:
Cool, so I'm ok to try to integrate this?

sormuras:
I'll sponsor this after

- #263

is integrated.

PR #268, opened 2025-06-24
7904045: Remove the macOS version check from jtreg

jaikiran:
Can I please get a review of this change which proposes to remove the macOS version check in jtreg tool? As noted in https://bugs.openjdk.org/browse/CODETOOLS-7904045, the current check results in jtreg exiting the process if it finds a mismatch in the macOS version reported by `sw_vers` tool and the one reported by Java's `os.version` system property. This makes it impossible to launch jtreg to run any tests on macOS 26 (currently in Beta). Even a fix in the JDK (through https://bugs.openjdk.org/browse/JDK-8359830) won't be enough because older versions of the JDK that could be used to launch the jtreg process will still cause this check to fail in jtreg.

The check itself isn't necessary within jtreg and that version check should instead reside as a regression test within the JDK repo. In fact there's already a regression test in JDK which catches these issues related to incorrect os.version reporting. Having this check in jtreg (that too only for macOS) prevents the tool from being used on these newer versions of macOS. In fact, jtreg itself cannot be built (with JDK 17) on macOS 26 (beta) due to this check.

The commit in this PR removes the check. I have run this change against macOS 26 (beta) where the tool would previously exit. The self-tests now all pass on that host:

```
All (     234) selected tests completed successfully
```

---------
### Progress

### Issue
 * [CODETOOLS-7904045](https://bugs.openjdk.org/browse/CODETOOLS-7904045): Remove the macOS version check from jtreg (**Enhancement** - P3)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/268/head:pull/268` \
`$ git checkout pull/268`

`$ git checkout pull/268` \
`$ git pull https://git.openjdk.org/jtreg.git pull/268/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 268`

`$ git pr show -t 268`

</details>

<a href="https://git.openjdk.org/jtreg/pull/268.diff">https://git.openjdk.org/jtreg/pull/268.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/268#issuecomment-2999225872)
</details>


sormuras:
The check was introduced in commit https://github.com/openjdk/jtreg/commit/84ac040e6e9cc281a5bc19a87f5b082e9febb560 without much reasoning recorded in https://bugs.openjdk.org/browse/CODETOOLS-7903088. As long as the OS version is still recorded when running `jtreg` for troubleshooting issues, I am fine with removin the check.

jaikiran:
> As long as the OS version is still recorded when running jtreg for troubleshooting issues

That's correct, no changes in that area and jtreg will continue to report the OS version in the relevant places like the jtr file of the test.

jaikiran:
Thank you Christian for the reivew. I'll go ahead and integrate this later in the day today.

PR #288, opened 2025-09-08
7904077: IDEA plugin: Plugin crashes when using newest Ant plug

JornVernee:
The jtreg plugins calls some methods from the ant plugin configuration classes. In the newest version of the ant plugin, one of these methods has been removed:

https://github.com/JetBrains/intellij-community/commit/265fee5a6e051d9d8f49403d48c5694a557e522c

This causes the jtreg plugin to crash with a NoSuchMethodError, and running tests is no longer possible. 

The fix proposed by this patch is to use the `Project` instance we get from the jtreg configuration instance. They should point to the same object.


JornVernee:
Thanks for the review!

/integrate

PR #291, opened 2025-09-24
7904084: Unset javaOpts and jtregOpts in jtreg.sh

sendaoYan:
Hi all,

In jtreg.sh it will combine all the java options and jtreg options in the `for i in "$@"` loop. SHELL use javaOpts and jtregOpts as global virable by default, this means jtreg.sh will receive the value of javaOpts and jtregOpts outside. I think this unexpected.
To avoid this siutation, I think we should unset javaOpts and jtregOpts before use it.

### DIFF
--- a/src/share/bin/jtreg.sh
+++ b/src/share/bin/jtreg.sh
@@ -162,6 +162,8 @@ DUALCASE=1  # for MKS: make case statement case-sensitive (6709498)
 saveIFS="$IFS"
 nl='
 '
+unset javaOpts
+unset jtregOpts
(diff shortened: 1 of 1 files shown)


sendaoYan:
Thanks for the reviews @sormuras 

Does this PR need second reviewer, or may I integrate this PR now.

liach (approved):
Clearing variable before loop makes sense.

sormuras:
> [...] may I integrate this PR now.

Yes. Happy to sponsor afterwards.

PR #318, opened 2026-04-09
7904176: JTReg should find inherited main methods

sormuras:
Please review this change fixing a regression introduced in https://bugs.openjdk.org/browse/CODETOOLS-7904141 - which failed to include inherited `main` methods from superclasses and implemeneted interface.

This PR also adds two related self-tests and updates expected summary output lines.

---------
### Progress

### Issue
 * [CODETOOLS-7904176](https://bugs.openjdk.org/browse/CODETOOLS-7904176): JTReg should find inherited main methods (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/318/head:pull/318` \
`$ git checkout pull/318`

`$ git checkout pull/318` \
`$ git pull https://git.openjdk.org/jtreg.git pull/318/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 318`

`$ git pr show -t 318`

</details>

<a href="https://git.openjdk.org/jtreg/pull/318.diff">https://git.openjdk.org/jtreg/pull/318.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/318#issuecomment-4216382728)
</details>

### DIFF
--- a/test/compactsourcefile/MainTreat.java
+++ b/test/compactsourcefile/MainTreat.java
@@ -1,0 +1,28 @@
+/*
+ * Copyright (c) 2026, Oracle and/or its affiliates. All rights reserved.
+ * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
+ *
+ * This code is free software; you can redistribute it and/or modify it
+ * under the terms of the GNU General Public License version 2 only, as
+ * published by the Free Software Foundation.
+ *
+ * This code is distributed in the hope that it will be useful, but WITHOUT
+ * ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or
+ * FITNESS FOR A PARTICULAR PURPOSE.  See the GNU General Public License
+ * version 2 for more details (a copy is included in the LICENSE file that
+ * accompanied this code).
+ *
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+interface MainTreat {
+    static void main() {
+        IO.println(getClass());
+    }
+}
--- a/test/compactsourcefile/MainTreat.java
+++ b/test/compactsourcefile/MainTreat.java
@@ -22,7 +22,7 @@
  */
 
 interface MainTreat {
-    static void main() {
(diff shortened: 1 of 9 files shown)


test/compactsourcefile/MainTreat.java
```diff
@@ -0,0 +1,28 @@
...
+ * You should have received a copy of the GNU General Public License version
+ * 2 along with this work; if not, write to the Free Software Foundation,
+ * Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA.
+ *
+ * Please contact Oracle, 500 Oracle Parkway, Redwood Shores, CA 94065 USA
+ * or visit www.oracle.com if you need additional information or have any
+ * questions.
+ */
+
+interface MainTreat {
```
jaikiran: Nit - was this meant to be `MainTrait`?

jaikiran:
Hello Christian, out of curiosity, how was this issue uncovered? One problem I keep having with the changes related to "find eligible main() method" code, both here and in the JDK repo is that it's not straightforward to remember these rules. I am reading the JEP on instance main methods and I'll review this PR shortly to see if this works as expected and if anymore cases need to be covered.

test/compactsourcefile/MainTreat.java
sormuras: Yes. Changing the names accordingly.

sormuras:
An initial [run](https://github.com/sormuras/jdk/actions/runs/24187228987) of the JDK build using a specific (close to HEAD) revision of jtreg showed that a few main tests failed with "no such method".

jaikiran:
Hello Christian,

> An initial [run](https://github.com/sormuras/jdk/actions/runs/24187228987) of the JDK build using a specific (close to HEAD) revision of jtreg showed that a few main tests failed with "no such method".

Our initial goal when introducing the ability to launch the modern main methods for jtreg tests was to try and avoid any impact on existing tests even if they were running recent JDK versions. Unfortunately it looks like even with those efforts, we ran into issues like the current one. The fix you have here I think is OK, but like I noted previously, this main() method selection isn't straightforward any longer and I won't be surprised if we run into more issues. I think we should file a separate enhancement request in jtreg (I can do that) which will introduce a flag which will allow projects that use jtreg to turn off this feature of detecting and using modern main methods. By default, we continue to try and find such main methods for newer JDK versions, but I think projects should be able to disable it and we should then just fallback to finding a public static void main like previously. I think that would provide a way to reduce the impact of this change.
As for this current change, it looks OK to me and if it helps us get past the current set of failures, then I think we should go ahead with it.

jaikiran (approved):

sormuras:
Thanks for the review, Jaikiran.

/integrate

PR #322, opened 2026-05-12
7904183: Some "main" action jtreg tests fail to launch after CODETOOLS-7904141

jaikiran:
Can I please get a review of this change which proposes to address the issue noted in https://bugs.openjdk.org/browse/CODETOOLS-7904183?

As noted in that issue, in a recent enhancement in jtreg 8.3, we introduced the ability to allow jtreg tests to run application main methods that follow the newer rules specified in JEP-512 https://openjdk.org/jeps/512 ("Compact Source Files and Instance Main Methods"). A major part of that feature involves identifying the "right" main method for a class.

Our implementation of that feature in jtreg was done through https://bugs.openjdk.org/browse/CODETOOLS-7904141. In that implementation we were cautious not to enable this feature for Java runtimes lesser than Java 25 (Java 25 delivered JEP-512). That way, the compatibility impact of this change would be minimal. Even so, we have now run into a few issues where some existing tests when running on Java 25 (or higher) may run into failures like the one reported in CODETOOLS-7904183. The underlying cause of this failure is due to how Java reflection may throw exceptions for unused class references when trying to lookup methods on a class. Details of that are noted in https://bugs.openjdk.org/browse/JDK-8351188. 

The commit in this PR addresses this failure by fixing the implementation of how we detect the main method. This implementation now closely matches with what the JDK's java launcher does when trying to lookup application main method. With this newer implementation, we prioritize the presence of a "public static void main(String[])" before doing additional reflective lookups on the class. That way, the vast majority of existing classes which have been coded to older style of main methods, continue to function normally.

A new self test has been introduced which reproduces the failure and verifies the fix.

One thing that these recent issues have shown is that, it might be better to make this feature an opt-in even when we detect Java 25 or higher runtime environments. The application main method detection logic has run into subtle failures even in the JDK, so I think it's a bit too early to enable this by default in jtreg. A separate enhancement request is being worked upon to make this a opt-in.

---------

---------
### Progress

### Issue
 * [CODETOOLS-7904183](https://bugs.openjdk.org/browse/CODETOOLS-7904183): Some "main" action jtreg tests fail to launch after CODETOOLS-7904141 (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/322/head:pull/322` \
`$ git checkout pull/322`

`$ git checkout pull/322` \
`$ git pull https://git.openjdk.org/jtreg.git pull/322/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 322`

`$ git pr show -t 322`

</details>

<a href="https://git.openjdk.org/jtreg/pull/322.diff">https://git.openjdk.org/jtreg/pull/322.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/322#issuecomment-4430913060)
</details>

### DIFF
--- a/src/share/classes/com/sun/javatest/regtest/agent/MainActionHelper.java
+++ b/src/share/classes/com/sun/javatest/regtest/agent/MainActionHelper.java
@@ -188,7 +188,7 @@ public class MainActionHelper extends ActionHelper {
                 argTypes = new Class<?>[] { String[].class };
                 methodArgs = new Object[] { classArgsArray };
                 method = MainMethodHelper.isModernMainSupported()
-                        ? MainMethodHelper.findMainMethod(c)
+                        ? MainMethodHelper.requireMainMethod(c)
                         : c.getMethod("main", argTypes);
             }
 
--- a/src/share/classes/com/sun/javatest/regtest/agent/MainMethodHelper.java
+++ b/src/share/classes/com/sun/javatest/regtest/agent/MainMethodHelper.java
@@ -29,9 +29,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
-import java.util.List;
-import java.util.stream.Collectors;
-import java.util.stream.Stream;
+import java.util.Objects;
 
 /**
  * Java program entry-point finder and runner.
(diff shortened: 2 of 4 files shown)


jaikiran:
The failures look related and I'm addressing them. It looks like these tests only run if a specific JDK25_HOME environment variable is set.

sormuras:
> The failures look related and I'm addressing them. It looks like these tests only run if a specific JDK25_HOME environment variable is set.

Yes. `CompactSourceFile25` uses `-jdk:/opt/hostedtoolcache/Java_Oracle_jdk/25/x64` to ensure it is run on JDK 25.

jaikiran:
I've updated the PR to address the test failures. The updated changes try to be as close as possible to what the JDK's code does when it comes to identifying a main method. With these updates changes, the existing and the new tests now pass.

sormuras (approved):

jaikiran:
Thank you Christian for the review.

PR #323, opened 2026-05-13
7904184: Introduce a jtreg property which allows for compact source and instance main methods feature to be enabled/disabled

jaikiran:
Can I please get a review of this change which addresses https://bugs.openjdk.org/browse/CODETOOLS-7904184?

A new jtreg property `allowCompactSourceInstanceMainMethod` has been introduced to allow projects to optionally disable the compact source and instance main method detection feature that we recently introduced in jtreg through https://bugs.openjdk.org/browse/CODETOOLS-7904141. This should allow projects to optionally disable this feature for specific tests if there are any unforeseen issues in the implementation of CODETOOLS-7904141 and at the same this feature will start getting usage in newer versions of the JDK.

I haven't documented the user facing `allowCompactSourceInstanceMainMethod` property, and that's intentional. I expect this property to be removed once we have seen enough usage of jtreg with this feature enabled (by default).

Jon @jonathan-gibbons, if you have a few moments, I would like your inputs on how this is implemented in this PR. The `main` action (through `MainAction` class) needs to propagate the "allow modern main methods" detail to the agent VM or the newly launched other VM. That detail is derived from TEST.ROOT/TEST.properties file. In the current proposed implementation, once the jtreg process has parsed that detail for the test, it propagates this detail to the agent VM and/or the other VM, through a new internal system property `test.allowModernMain`. That system property value is then read by the other process (either an agent or a new JVM) before the main method detection logic is invoked. I haven't found any other way, other than passing a system property, to achieve this co-ordination. Are there other ways to do this?

---------

---------
### Progress

### Issue
 * [CODETOOLS-7904184](https://bugs.openjdk.org/browse/CODETOOLS-7904184): Introduce a jtreg property which allows for compact source and instance main methods feature to be enabled/disabled (**Enhancement** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/323/head:pull/323` \
`$ git checkout pull/323`

`$ git checkout pull/323` \
`$ git pull https://git.openjdk.org/jtreg.git pull/323/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 323`

`$ git pr show -t 323`

</details>

<a href="https://git.openjdk.org/jtreg/pull/323.diff">https://git.openjdk.org/jtreg/pull/323.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/323#issuecomment-4441905175)
</details>

### DIFF
--- a/src/share/classes/com/sun/javatest/regtest/agent/MainActionHelper.java
+++ b/src/share/classes/com/sun/javatest/regtest/agent/MainActionHelper.java
@@ -134,7 +134,7 @@ public class MainActionHelper extends ActionHelper {
         SaveState saved = new SaveState();
 
         Properties p = System.getProperties();
-        boolean allowModernMain = false;
+        boolean allowModernMain = true; // enabled by default
         for (Map.Entry<String, String> e : props.entrySet()) {
             String name = e.getKey();
             String value = e.getValue();
--- a/src/share/classes/com/sun/javatest/regtest/agent/MainActionHelper.java
+++ b/src/share/classes/com/sun/javatest/regtest/agent/MainActionHelper.java
@@ -134,12 +134,15 @@ public class MainActionHelper extends ActionHelper {
         SaveState saved = new SaveState();
 
         Properties p = System.getProperties();
+        boolean allowModernMain = false;
         for (Map.Entry<String, String> e : props.entrySet()) {
             String name = e.getKey();
             String value = e.getValue();
             if (name.equals("test.class.path.prefix")) {
                 SearchPath cp = new SearchPath(value, System.getProperty("java.class.path"));
                 p.put("java.class.path", cp.toString());
+            } else if (name.equals("test.allowModernMain")) {
+                allowModernMain = Boolean.parseBoolean(value);
             } else {
                 p.put(e.getKey(), e.getValue());
             }
@@ -187,7 +190,7 @@ public class MainActionHelper extends ActionHelper {
                 // Normal case: marker interface not found; use standard main method
                 argTypes = new Class<?>[] { String[].class };
                 methodArgs = new Object[] { classArgsArray };
-                method = MainMethodHelper.isModernMainSupported()
+                method = allowModernMain && MainMethodHelper.isModernMainSupported()
                         ? MainMethodHelper.requireMainMethod(c)
                         : c.getMethod("main", argTypes);
             }
--- a/src/share/classes/com/sun/javatest/regtest/agent/MainWrapper.java
+++ b/src/share/classes/com/sun/javatest/regtest/agent/MainWrapper.java
@@ -134,8 +134,9 @@ public class MainWrapper
 
                 // RUN JAVA PROGRAM
                 Class<?> c = Class.forName(className, false, cl);
-                boolean allowModernMain = Boolean.getBoolean("test.allowModernMain");
-                if (allowModernMain && MainMethodHelper.isModernMainSupported()) {
+                // enabled by default
+                String allowModernMain = System.getProperty("test.allowModernMain", "true");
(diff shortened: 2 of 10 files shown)


jaikiran:
After discussing more with others and based on those inputs, I have updated the PR to only introduce a new jtreg property which allows the compact source and instance main method detection feature to be disabled (or enabled). By default the feature stays enabled. We decided that the current implementation is good enough to enable it by default and start getting coverage for it. Any unforeseen issues can be addressed by setting `allowCompactSourceInstanceMainMethod` to `false` and reporting those issues to jtreg.

A new self test has been introduced to verify that `allowCompactSourceInstanceMainMethod=false` works as expected. I've updated the issue title as well as the PR description to match the latest update.

jonathan-gibbons:
> I haven't documented the user facing allowCompactSourceInstanceMainMethod property, and that's intentional. I expect this property to be removed once we have seen enough usage of jtreg with this feature enabled (by default).

I would have expected such a property to be formally specified and defined in `TEST.ROOT` or `TEST.properties`.
I would expect the feature to be disabled by default, for compatibility, and only enabled in test suits or parts of a test suite where it is know that the use of the feature is valid. In other words, it should be disabled by default in those parts of a test suite where there may be false positive use of the feature.

It's been a while since I looked at the code, but IIRC there were ways to pass values to an agent, either when the agent was started, via command-line args, or on a per-test basis (perhaps using the _.jta_ file.)

Although it might potentially duplicate some work, you might also be able to do more analysis in the `jtreg` `MainAction` class. Presumably there is at least some checking there that there is a suitable `main` method.

Finally, given that there seem to be issues with the overall feature in both JDK and jtreg, I hope there is a comprehensive (and growing) set of regression tests for the feature, in the JDK test suite and the jtreg test suite.

sormuras (approved):

jaikiran:
Thank you Jon for these inputs.

> I would have expected such a property to be formally specified and defined in TEST.ROOT or TEST.properties.

Done, I've added the documentation for this property in the TEST.ROOT documentation where the other properties are listed.

> I would expect the feature to be disabled by default, for compatibility, and only enabled in test suits or parts of a test suite where it is know that the use of the feature is valid. In other words, it should be disabled by default in those parts of a test suite where there may be false positive use of the feature.

That was my initial thought as well as the initial implementation in this PR. I discussed it internally with others and their suggestion has been that since we enable this feature only when the test JDK is Java 25 or higher and since with this PR there's now a way to disable this feature if necessary, it would be good to have it enabled by default (for versions Java 25 or higher). I decided to follow that suggestion.

> Although it might potentially duplicate some work, you might also be able to do more analysis in the jtreg MainAction class. Presumably there is at least some checking there that there is a suitable main method.

Right, the `MainAction` is where we introduced these checks to determine whether or not to allow this feature and detect the newer style main methods.

> It's been a while since I looked at the code, but IIRC there were ways to pass values to an agent, either when the agent was started, via command-line args, or on a per-test basis (perhaps using the .jta file.)

I experimented with this the past few days. I have to admit I haven't been able to fully focus on this part, so those experiments haven't been conclusive (or clean). Given that this is an internal detail of jtreg and that a newer release of jtreg is being planned soon, I'll go ahead with what I've in this PR for now and come back to this at a later time when I am less distracted.

> Finally, given that there seem to be issues with the overall feature in both JDK and jtreg, I hope there is a comprehensive (and growing) set of regression tests for the feature, in the JDK test suite and the jtreg test suite.

A while back, Christian added self tests within jtreg to cover several of these main() method use cases. Plus the JDK mainline repo too has seen increased test coverage for compact source and instance main method feature. So yes, we have a growing set of regression tests in both these repos.

sormuras (approved):
Thanks for the PR and addressing review comments, Jaikiran.

PR #326, opened 2026-07-28
7904199: On MSys2, jtreg defaults to using wsl to run shell tests

raneashay:
Before this patch, jtreg chose to execute shell tests using WSL when
Cygwin was unavailable, but this breaks when jtreg is run using MSys,
since jtreg then incorrectly concludes that it needs to use WSL, even
though the "sh" binary is available under MSys.  This patch fixes the
problem so that jtreg chooses to use WSL only when _both_ Cygwin and
MSys are unavailable (in addition to checking whether WSL is detected).

---------

---------
### Progress
- \[x\] Change must be properly reviewed (1 review required, with at least 1 [Reviewer](https://openjdk.org/bylaws#reviewer))
- \[x\] Change must not contain extraneous whitespace
- \[x\] Commit message must refer to an issue

### Issue
 * [CODETOOLS-7904199](https://bugs.openjdk.org/browse/CODETOOLS-7904199): On MSys2, jtreg defaults to using wsl to run shell tests (**Bug** - P4)

### Reviewers

### Reviewing

Checkout this PR locally: \
`$ git fetch https://git.openjdk.org/jtreg.git pull/326/head:pull/326` \
`$ git checkout pull/326`

`$ git checkout pull/326` \
`$ git pull https://git.openjdk.org/jtreg.git pull/326/head`

</details>

Checkout this PR locally: \
`$ git pr checkout 326`

`$ git pr show -t 326`

</details>

<a href="https://git.openjdk.org/jtreg/pull/326.diff">https://git.openjdk.org/jtreg/pull/326.diff</a>

</details>

[Link to Webrev Comment](https://git.openjdk.org/jtreg/pull/326#issuecomment-5110326170)
</details>

### DIFF
--- a/src/share/classes/com/sun/javatest/regtest/tool/Tool.java
+++ b/src/share/classes/com/sun/javatest/regtest/tool/Tool.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 1997, 2025, Oracle and/or its affiliates. All rights reserved.
+ * Copyright (c) 1997, 2026, Oracle and/or its affiliates. All rights reserved.
  * DO NOT ALTER OR REMOVE COPYRIGHT NOTICES OR THIS FILE HEADER.
  *
  * This code is free software; you can redistribute it and/or modify it
@@ -2338,8 +2338,8 @@ public class Tool {
     }
 
     /**
-     * Returns whether MSys is available, by examining whether the `MSYSTEM`
-     * environment variable is non-empty.
+     * Returns whether MSys is available, by examining whether the
+     * {@code MSYSTEM} environment variable is non-empty.
      */
     private boolean isMSysDetected() {
         if (isWindows()) {
--- a/src/share/classes/com/sun/javatest/regtest/tool/Tool.java
+++ b/src/share/classes/com/sun/javatest/regtest/tool/Tool.java
@@ -682,6 +682,9 @@ public class Tool {
                 if (isCygwinDetected()) {
                     out.println(i18n.getString("main.warn.wsl.specified.found.cygwin"));
                 }
+                if (isMSysDetected()) {
+                    out.println(i18n.getString("main.warn.wsl.specified.found.msys"));
+                }
                 useWindowsSubsystemForLinux = true;
             }
         },
@@ -1197,7 +1200,7 @@ public class Tool {
             if (useWindowsSubsystemForLinux == null) {
                 // The test for Cygwin is more specific than the test for WSL,
                 // and so we give priority to Cygwin if both are detected.
-                useWindowsSubsystemForLinux = !isCygwinDetected() && isWindowsSubsystemForLinuxDetected();
+                useWindowsSubsystemForLinux = !isCygwinDetected() && !isMSysDetected() && isWindowsSubsystemForLinuxDetected();
             }
         } else {
             useWindowsSubsystemForLinux = false;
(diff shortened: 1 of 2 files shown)


jaikiran (commented):
Hello Ashay, the changes look reasonable to me. Each of these 2 files will need a copyright year update.

Have your run the jtreg self tests locally (on a Windows instance) to verify that existing self tests pass after this change?

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
```diff
@@ -2334,6 +2337,19 @@ private boolean isCygwinDetected() {
         }
     }
 
+    /**
+     * Returns whether MSys is available, by examining whether the `MSYSTEM`
```
jaikiran: Nit the use of backtick isn't necessary here. Instead use `{@code MSYSTEM}`.

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
```diff
@@ -2334,6 +2337,19 @@ private boolean isCygwinDetected() {
...
+     * environment variable is non-empty.
+     */
+    private boolean isMSysDetected() {
+        if (isWindows()) {
+            String msystem = System.getenv("MSYSTEM");
```
jaikiran, line 2346: Is this environment variable documented in some official doc? I don't expect it to be linked as comment in the code, but it would be good to know if any such authoritative document exists. I did a extremely brief search and haven't found anything relevant.

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
raneashay: Thanks, I wasn't aware of this annotation!  Fixed in the most recent commit.

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
raneashay: I used this page https://www.msys2.org/docs/environments as reference, based on which I surmised that `MSYSTEM` is set to one of `MSYS`, `UCRT64`, `CLANG64`, `CLANGARM64`, `CLANG32`, `MINGW32`, or `MINGW64`.  I validated this on a few machines running either Windows/x64 or Windows/ARM64.

raneashay:
> Each of these 2 files will need a copyright year update.

Dang, sorry, I keep forgetting to update the year.  Fixed in the most recent commit.

> Have your run the jtreg self tests locally?

Yes, I don't see any failures, although there are couple of caveats: (1) I had to run the tests under Cygwin (I haven't triaged the key problem yet) and (2) I had to apply several fixes to the tests to handle the CRLF line endings and the differences in file paths (forward versus backward slashes).  I'll file a separate JBS issue (EDIT: see https://bugs.openjdk.org/browse/CODETOOLS-7904200) and PR for the fixes to the tests.

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
jaikiran: For additional reference, it's one of the standard javadoc tags. The reference for those is available here https://docs.oracle.com/en/java/javase/26/docs/specs/javadoc/doc-comment-spec.html#standard-tags

src/share/classes/com/sun/javatest/regtest/tool/Tool.java
jaikiran: Thank you, I think that reference and validation is good enough.

jaikiran:
Hello Ashay,

> I had to run the tests under Cygwin (I haven't triaged the key problem yet) 

Do  you mean, even after the proposed changes in this PR, you weren't able to use MSys2 to run the self tests?

jaikiran (approved):
Thank you for the updates. This looks good to me.

raneashay:
> even after the proposed changes in this PR, you weren't able to use MSys2 to run the self tests?

Yes, indeed.  There's some issue with the test framework that causes the classpath to be set incorrectly.  Here's the error that I see:

```
Error: Unable to initialize main class Basic
Caused by: java.lang.NoClassDefFoundError: com/sun/javatest/Harness$Fault
make: *** [../test/basic/Basic.gmk:87: jtreg/build/test/Basic.othervm.ok] Error 1
```

Once I find the fix, I'll file a separate PR.

jaikiran:
> There's some issue with the test framework that causes the classpath to be set incorrectly.

jtreg (even during self tests) creates a `.jtr` file for each test. For the failing test too, somewhere under the build directory, there will be a `.jtr` file. In that file, you should find all the details, including the exact command containing that classpath and maybe that will provide enough hints.

